PR #1, opened 2018-10-05
initial outline with Russ Rutledge and Cedric Williams and Noah Cawley

gruetter:
/CC @cewilliams , @rrrutledge, @_utanapishtim

### DIFF
--- a/trusted-committer/outline.md
+++ b/trusted-committer/outline.md
@@ -1,0 +1,86 @@
+# Trusted Committer Learning Path Outline
+
+## Learning Goals
+
+- Definition of TC (why that name)
+- Understanding of the responsibilities of the Trusted Committer (TC)
+- Different ways to become TC
+- TC lifecycle (bringing on successors, sunsetting)
+- Useful skills/attitude/emotional maturity
+- TC antipatterns
+- Why is the TC role important for IS success
+
+## Target Audience
+
+- Current TCs
+- People who work/interface with TCs 
+    - Contributors
+    - Potential future TCs
+    - Managers of TCs
+
+## Outline
+
+### Definition
+
+- Glossary
+    - Contributor/Guest
+        - Responsible for "fitting in"/conforming to the host team and it's
+          expectations/processes.
+    - Product Owner (distinction btw. Guest and Host PO)
+    - TC
+        - TC has rights to push code closer to production than a contributor
+        - also known as _Host_
+        - has privileges that an admin or Apache committer would have but it is
+          a job, irrespective of the permissions. 
+        - Maintainer/Owner (conflict with _maintainer_ role in GitHub)
+        - Committer is not an IS term. Apache uses this term to denote a tech
+          leadership role (TC++)
+        - The TC role is defined by its responsibilities, rather than by its
+          privileges.
+
+- at PayPal, the discussion surrounding the name of the role was very
+  challenging.  i. e.: the name matters.
+
+### Responsibilities
+
+**Tech oriented**
+- Keep SW development running smoothly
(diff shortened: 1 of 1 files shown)


trusted-committer/outline.md
```diff
@@ -0,0 +1,86 @@
...
+    - Product Owner (distinction btw. Guest and Host PO)
+    - TC
+        - TC has rights to push code closer to production than a contributor
+        - also known as _Host_
+        - has privileges that an admin or Apache committer would have but it is
+          a job, irrespective of the permissions. 
+        - Maintainer/Owner (conflict with _maintainer_ role in GitHub)
+        - Committer is not an IS term. Apache uses this term to denote a tech
+          leadership role (TC++)
+        - The TC role is defined by its responsibilities, rather than by its
```
rrrutledge: Great quote!

trusted-committer/outline.md
```diff
@@ -0,0 +1,86 @@
...
+          privileges.
+
+- at PayPal, the discussion surrounding the name of the role was very
+  challenging.  i. e.: the name matters.
+
+### Responsibilities
+
+**Tech oriented**
```
rrrutledge: Maybe as a counterpoint state that TCs are _not_ responsible for writing guest contributions for them or rewriting contributions that come in not-to-standard.

trusted-committer/outline.md
```diff
@@ -0,0 +1,86 @@
...
+- 
+
+### Lifecycle
+
+- There might well be multiple TCs in one project
+- 
+
+## Ideas
+
+- Possibly two videos: one for being and one for becoming a TC
```
rrrutledge: Like this idea.

trusted-committer/outline.md
gruetter: I agree that they are not responsible for that, indeed. But they should not be passive on that front either. I have always viewed it as part of my job as TC to assist people in making their contributions conform to the required standards. Usually by way of conversations in PRs, sometimes including pseudocode or sketches of algos. I generally avoided rewriting SW for contributors, but it did happen more than once.

trusted-committer/outline.md
```diff
@@ -0,0 +1,86 @@
...
+- The contributor learning path (LP) video should lay out what a contributor
+  can expect from a/his TC
```
lenucksi: Sounds like a good idea. #37 should consider this while we're at it.

trusted-committer/outline.md
lenucksi: 👍

trusted-committer/outline.md
lenucksi: Actually, that's a nice idea. Did we do any of that (still need to go through all the Screenlight things)? Would we want to have small parts of it included in the contributor segment (#37) along the lines of "from contributor to trusted committer" - that's something that would likely work with open source projects, I'm not sure how well it works for all inner source projects, with the grass-roots inner source projects that should probably be fine.

trusted-committer/outline.md
```diff
@@ -0,0 +1,86 @@
...
+- Idea for a new LP: InnerSource careers
```
lenucksi: ...file under "ideas for a second learning path". There surely are quite a number of aspects that could be handled / presented.
I propose to have a open questions / further development / further research session+document after finishing this LP to preserve all the good things for a second / next one.

rrrutledge:
Sounds good to me.

trusted-committer/outline.md
lenucksi: #62 needs to know about this

rrrutledge: Thanks for adding this note.

PR #4, opened 2018-10-09
Create introduction/principles.md

rrrutledge:
This file holds the text to accompany the [InnerSource Principles](https://www.safaribooksonline.com/videos/introduction-to-innersource/9781492041504/9781492041504-video321610) video.
This accompanying text will appear as a part of the learning path (like [this example](https://www.safaribooksonline.com/learning-paths/learning-path-lean/9781491999738/9781491946527-/part01ch01.html)).

### DIFF
--- a/introduction/principles.md
+++ b/introduction/principles.md
@@ -1,0 +1,39 @@
+This file holds the text to accompany the [InnerSource Principles](https://www.safaribooksonline.com/videos/introduction-to-innersource/9781492041504/9781492041504-video321610) video.
+This accompanying text will appear as a part of the learning path (like [this example](https://www.safaribooksonline.com/learning-paths/learning-path-lean/9781491999738/9781491946527-/part01ch01.html)).
+
+# Inner Source Principles
+
+Every company, team, project, and individual is different.
+Because of that fact, the exact way that the concept of inner source works will vary from one situation to another.
+At its core, however, are two principles that form the bedrock of any successful instance of inner source.
+These principles have inspiration in successful open source projects and must be respected and present in order for inner source to achieve the benefits described earlier.
+
+The principles are:
+* **Prioritized Mentorship**
+* **Voluntary Code Contribution**
+
+Let's take a look at each of these principles in more detail.
+
+## Prioritized Mentorship
+
+_Mentorship_ from host team to guest team via trusted committers is a key aspect of inner source.
+Contributors on guest teams are upleveled so that they understand enough about the host team system to change it successfully.
+In the process of doing so, they come to better understand the host team system as a general consumer.
+They can use it more effectively and help other consumers to do so as well.
+
+It's critical that this mentorship for contributors is _Prioritized_ by the host team.
+The team strives to make trusted committer time available to mentor guest team contributors _at the time that they need it_ as opposed to when it is convenient.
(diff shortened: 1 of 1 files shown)


introduction/principles.md
```diff
@@ -0,0 +1,39 @@
...
+
+Every company, team, project, and individual is different.
+Because of that fact, the exact way that the concept of inner source works will vary from one situation to another.
+At its core, however, are two principles that form the bedrock of any successful instance of inner source.
+These principles have inspiration in successful open source projects and must be respected and present in order for inner source to achieve the benefits described earlier.
+
+The principles are:
+* **Prioritized Mentorship**
+* **Voluntary Code Contribution**
+
```
gruetter: And here comes my customary suggestion to include more principles. IMHO, InnerSource wouldn't work without

- **Openness**: projects need to be set up in a way that makes contributions possible in the first place)
- **Transparency**: work products, progress, decision making must be transparent in order to be able to work asynchronously). This is loosely related to **Prioritized Mentorship**, IMO.
- **Meritocracy**: those who _do_ have the say. This is somewhat related to **Prioritized Mentorship**, too.

KathrynKLewis: I agree w @gruetter points. As the reader, I don't automatically understand what "prioritized mentorship" means and if I try to infer a meaning I think it could be a benefit for the guest/contributor only. If possible, I would suggest reworking it to center around the principles that @gruetter mentioned. If we want to stick with two, here are my suggestions: 
1) Openness: projects are set up so that they can be widely accessed for reuse and ongoing improvements
2) Transparency: projects are oriented so that all progress, decision making and current state is visible and accessible

rrrutledge: Thanks for this feedback.  Maybe we can have a discussion about this point on a call.  The feedback here is significant enough to possibly require a video reshoot.

rrrutledge: I'm reading through the feedback again, and Im not convinced that these items are necessary on their own for inner source to work in general.  I think that your inner source effort will scale larger with them, but an individual team could incorporate inner source into their regular workflow without them.

rrrutledge: Had a discussion in [these notes](https://paypalflow.slack.com/archives/CARTU4XV2/p1548443862048400) and ended up wanting to re-write this article.

demianbrecht: There are actually a couple other points that I'd like to tack on from my (shameless plug) [blog post](https://engineering.salesforce.com/inner-sourcing-the-mechanics-c0b1421230fd):

* Automation: It's critical for any project to (at minimum) automated unit tests in place. Without that infrastructure, it makes it possible for bugs to slip through the cracks during PR reviews. This should also mention some relatively high and somewhat arbitrary threshold for test coverage.
* Documentation: I'm not sure whether or not this was intended to be included in openness, but I feel it's critical to any projects success at being either inner or open sourced. It might also be worthwhile to get into the difference between developer and user docs, the former being critical to inner sourcing.

rrrutledge: We think that these are elements of being (technically) open.

rrrutledge:
Closing in favor of the upcoming rewrite from @demianbrecht (https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/36).

PR #7, opened 2018-10-09
Create introduction/problems-solved.md

rrrutledge:
This file holds the text to accompany the [What Problems Does InnerSource Solve?](https://www.safaribooksonline.com/videos/introduction-to-innersource/9781492041504/9781492041504-video321607) video.
This accompanying text will appear as a part of the learning path (like [this example](https://www.safaribooksonline.com/learning-paths/learning-path-lean/9781491999738/9781491946527-/part01ch01.html)).

### DIFF
--- a/introduction/problems-solved.md
+++ b/introduction/problems-solved.md
@@ -3,8 +3,12 @@ This accompanying text will appear as a part of the learning path (like [this ex
 
 # What Problems Does InnerSource Solve?
 
-To illustrate the types of situations where inner source can help, imagine the following.
-Suppose two teams at the same company deliver separate pieces of software with one team's software depending on that of the other.
+Inner source encourages and rewards collaboration from anyone, regardless of their location in company organizational structure.
+This approach differs from the results seen in traditional organizations, where ideas and work tend to stay trapped within the boundaries of internal corporate hierarchy and the work and planning processes that align themselves to that hierarchy.
+Inner source provides a way for these ideas and work to break free of those traditional bounds to the benefit of all involved.
+Let's explore one situation that gives an example of this idea. 
+
+Imagine two teams at the same company deliver separate pieces of software with one team's software depending on that of the other.
 An example could be a user experience that depends on an API service to retrieve data for display.
 This situation is common in a large enterprise where a single team producing software may have dozens or hundreds of consumers.
 
@@ -18,7 +22,7 @@ For critical feature requests that are not prioritized for immediate work, the c
 1. **Workaround**. A consuming team that doesn't want to wait may do extra work somewhere else to compensate for the absence of their requested feature.
   This extra work may come as change in the consuming project.
   Alternatively, they may create a new project that meets their needs and replaces their use of all or part of the producing team's system (code/project duplication).
-  This strategy allows the consuming team to get the requested feature via their own efforts only (no hard dependencies). It comes with several drawbacks, though.
+  This strategy allows the consuming team to get the requested feature via their own efforts only. It comes with several drawbacks, though.
       
     1. Any work done by the consuming team remains unavailable to any other consumers with the same feature request.
     1. The consuming team has inadvertently signed up for the long-term burden of maintaining the newly-written code, which is not in the domain of their core team competency.
@@ -27,9 +31,10 @@ For critical feature requests that are not prioritized for immediate work, the c
 1. **Escalate**. The consuming team may not take "no" for an answer and, instead, advocate to someone in the producers' management hierarchy to influence (or force) the producing team to do the work.
 This option sounds attractive for the consuming team because they get the requested feature without doing the work to implement or maintain it.
 It is still a drag on the team, though, because it necessarily diverts some of their attention and work to the non-engineering task of escalation.
-Additionally, this option does not scale as there are so many times that a consumer can escalate feature requests before damaging their credibility.
-Escalate is similarly disruptive (even more so) for the producing team, who are taken out of their normal workflow and prioritization methods to deal with the escalated feature request.
+Additionally, this option does not scale as there are only so many times that a consumer can escalate feature requests before damaging their credibility.
+Escalation is similarly disruptive (even more so) for the producing team, who are taken out of their normal workflow and prioritization methods to deal with the escalated feature request.
 
 This discussion sets the stage for inner source.
 Inner source applies to the same type of situation where a consuming team is unable to get what needs via feature request.
 Inner source provides a way for the teams to gain the benefits of _wait it out_, _workaround_, and _escalate_ without the associated drawbacks.
+It also provides a general improvement to engineering culture as engineers have the chance to work with a wider variety of new technologies and people.
--- a/introduction/problems-solved.md
+++ b/introduction/problems-solved.md
@@ -3,9 +3,8 @@ This accompanying text will appear as a part of the learning path (like [this ex
 
 # What Problems Does InnerSource Solve?
 
-Inner source encourages and rewards collaboration from anyone, regardless of their location in company organizational structure.
-This approach differs from the results seen in traditional organizations, where ideas and work tend to stay trapped within the boundaries of internal corporate hierarchy and the work and planning processes that align themselves to that hierarchy.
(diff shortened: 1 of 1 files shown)


introduction/problems-solved.md
```diff
@@ -0,0 +1,35 @@
+This file holds the text to accompany the [What Problems Does InnerSource Solve?](https://www.safaribooksonline.com/videos/introduction-to-innersource/9781492041504/9781492041504-video321607) video.
+This accompanying text will appear as a part of the learning path (like [this example](https://www.safaribooksonline.com/learning-paths/learning-path-lean/9781491999738/9781491946527-/part01ch01.html)).
+
+# What Problems Does InnerSource Solve?
+
+To illustrate the types of situations where inner source can help, imagine the following.
```
gruetter: I feel that InnerSource solves problems above and beyond the situation you described in this article. On a more abstract level, it fosters collaboration and learning where both was not possible beforehand due to strong siloing. The situation you describe here is "merely" one symptom of that problem. How about we lead with a more general statement along the lines of what I mentioned above and then follow with the concrete example you describe here.

introduction/problems-solved.md
```diff
@@ -0,0 +1,35 @@
...
+For critical feature requests that are not prioritized for immediate work, the consuming team may commonly choose one of 3 options, each of which come with their own drawbacks.
+
+1. **Wait it out**. The consuming team may do nothing and limp along without the requested functionality.
+  This option requires the least amount of work on their side.
+  Depending on the benefit of the feature request, waiting might be just fine.
+  However, it may come with real amounts of pain, especially if the requested functionalty is never delivered.
+1. **Workaround**. A consuming team that doesn't want to wait may do extra work somewhere else to compensate for the absence of their requested feature.
+  This extra work may come as change in the consuming project.
+  Alternatively, they may create a new project that meets their needs and replaces their use of all or part of the producing team's system (code/project duplication).
+  This strategy allows the consuming team to get the requested feature via their own efforts only (no hard dependencies). It comes with several drawbacks, though.
```
gruetter: > (no hard dependencies)

Just wondering: what would be a _soft_ dependency?

introduction/problems-solved.md
```diff
@@ -0,0 +1,35 @@
...
+      
+    1. Any work done by the consuming team remains unavailable to any other consumers with the same feature request.
+    1. The consuming team has inadvertently signed up for the long-term burden of maintaining the newly-written code, which is not in the domain of their core team competency.
+    1. The company as-a-whole acquires duplicate projects and code in the same problem space.
+      
+1. **Escalate**. The consuming team may not take "no" for an answer and, instead, advocate to someone in the producers' management hierarchy to influence (or force) the producing team to do the work.
+This option sounds attractive for the consuming team because they get the requested feature without doing the work to implement or maintain it.
+It is still a drag on the team, though, because it necessarily diverts some of their attention and work to the non-engineering task of escalation.
+Additionally, this option does not scale as there are so many times that a consumer can escalate feature requests before damaging their credibility.
```
gruetter: I'm not a native speaker, of course, but shouldn't we `s/there are so many times/there are _only_ so many times`?

introduction/problems-solved.md
```diff
@@ -0,0 +1,35 @@
...
+Escalate is similarly disruptive (even more so) for the producing team, who are taken out of their normal workflow and prioritization methods to deal with the escalated feature request.
```
gruetter: s/Escalate/Escalation

introduction/problems-solved.md
```diff
@@ -0,0 +1,35 @@
...
+
+This discussion sets the stage for inner source.
```
gruetter: Maybe here would be a good point where we could mention other problems that are solved very briefly, such as:

- lack of mentoring (often happens in PRs in InnerSource)
- lack of infrastructure and empowering to collaborate across silos
- lack of infrastructure to work asynchronously
- absence of technical career path (vs. meritocracy in InnerSource)

introduction/problems-solved.md
```diff
@@ -0,0 +1,35 @@
+This file holds the text to accompany the [What Problems Does InnerSource Solve?](https://www.safaribooksonline.com/videos/introduction-to-innersource/9781492041504/9781492041504-video321607) video.
```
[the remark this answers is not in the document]
rrrutledge: Thanks.

introduction/problems-solved.md
rrrutledge: Thanks for these ideas.  These points sounds more like benefits to me as each one is phrased as the lack of a benefit rather than the presence of a problem.  Maybe we can call them out in [the benefits article](https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/5#issuecomment-454507604)?

rrrutledge: Added a short sentence summarizing the idea that there are other benefits to inner source besides what we've listed here.

introduction/problems-solved.md
rrrutledge: Sounds good.

introduction/problems-solved.md
rrrutledge: Apparently you are a better speaker than I am 😜

introduction/problems-solved.md
rrrutledge: You're right, that term is not well-defined.  Removed it.

introduction/problems-solved.md
rrrutledge: Great idea!  Added an extra paragraph at the beginning.

introduction/problems-solved.md
```diff
@@ -0,0 +1,40 @@
...
```
KathrynKLewis: Suggest rewording intro graf to be less wordy: 

Inner source encourages and rewards collaboration and code reuse with anyone, regardless of their position in a company's organizational structure. This approach differs what is seen in traditional organizations where ideas and work product tend to stay trapped within the boundaries of the internal corporate hierarchy and its silos.

introduction/problems-solved.md
KathrynKLewis: I agree w Georg's comments. I think this intro should more clearly describe other problems that inner source solves so that we are capturing the reader's interest and curiosity to learn more.  One element that I think ought to be mentioned is as follows: 
Inner source also allows teams to work more efficiently by taking advantage of other software already built by teams across their organization. Team members do not have to repeatedly reinvent the wheel or solve problems that have already been solved because they can access the related or similar work that other teams have already done. Teams can then focus on higher value problems by leveraging their colleagues' work product.

rrrutledge: Updated the final paragraph - does that incorporate those critical ideas?

introduction/problems-solved.md
rrrutledge: I love reducing words!

gruetter: love this, @KathrynKLewis !

introduction/problems-solved.md
gruetter: I still think that _mentoring_ and _learning_ should be mentioned explicitly. The first sentence in the final paragraph focuses on _exposure_ to new tech and people but doesn't mention mentoring at all. That is, in my experience, what sets InnerSource apart from traditional projects, where you are also exposed to new people and tech, when you switch projects from time to time. Just my 2 c

rrrutledge: Done.

introduction/problems-solved.md
```diff
@@ -38,5 +38,5 @@ Inner source applies to the same type of situation where a consuming team is una
 Inner source provides a way for the teams to gain the benefits of _wait it out_, _workaround_, and _escalate_ without the associated drawbacks.
 
 Inner source also provides a general improvement to engineering culture as engineers have the chance to work with a wider variety of new technologies and people.
-It facilitates sharing of ideas and solutions across organizational silos.
+Developers mentor and learn from one other as they share ideas and solutions across organizational silos.
```
lenucksi: Should this not be 'learn from one **an**other'?
Other than that, 👍.

rrrutledge:
We wrote the last paragraph together after the video was shot.  Perhaps when we re-shoot the `problems solved` video we can re-shoot this one, too.

PR #18, opened 2018-10-27
Incorporated feedback from audio rehearsal. 

gruetter:
This PR adds more improvements based on super valuable feedback and suggestions from Loren Sanz. Thanks again for that, Loren! I've also added brief summaries at the end of each section to the videos for each a bit more well-rounded.


gruetter:
@guywmartin : I just saw that I have some uncommitted changes to the article. (#23)

PR #22, opened 2018-11-17
Cleaned up some general grammar and missing words.

guywmartin:


lenucksi:
The suggestions look good, I tried to apply them to the current master, however found that they differ quite a lot now and thus make the patch hard to apply.

I tried to apply it and there are a few sections that differ from a writing or content perspective, making a non-discussion merge somewhat unfeasible. How should we proceed with this PR?

rrrutledge:
I think that we should just leave this PR closed as I think Georg was able to incorporate the gist of these suggestions from Guy before filming.

lenucksi:
> I think that we should just leave this PR closed as I think Georg was able to incorporate the gist of these suggestions from Guy before filming.

To leave it closed we first have to actually close it - it's marked green/open here. ;)
Until further notice I'll consider its contents done / integrated though.

rrrutledge:
Oh, you're right, the _GitHub_ issue was closed without closing this PR.  @guywmartin, @gruetter, let me know if I'm wrong but if I remember right you all were able to get in the relevant thoughts from this PR already, so closing now.  Feel free to reopen if I'm mistaken.

rrrutledge:
Thanks for catching this stale PR, @lenucksi

guywmartin:
Thanks @rrrutledge - I think you're correct - @gruetter incorporated my suggestions before filming.

PR #24, opened 2018-11-20
Update README.md

rrrutledge:
Add explanation of what is the learning path and how to get involved.  Previously only given verbally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This repository holds the backing and planning material for the videos and artic
 
 ## Content Organization
 
-There are (roughly) four sections to the learning path:
+For now, there are four sections to the learning path each covering a specific topic:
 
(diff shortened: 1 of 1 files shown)


rrrutledge:
> * i.e. state that the readme is current as of X date with

I think that this is covered with _GitHub_ already displaying the date of last commit at the top of the page.

rrrutledge:
> * some info on how the project will progress over time - when you expect videos to be complete and some statement about how you want to handle the future expansion of the learning path (maybe it will just be organic via the slack channel dialog).

Added some language directing all other inquiries to _Slack_ (basically we don't know what will happen in the future, yet).

rrrutledge:
Thanks for the review, @KathrynKLewis and @maxcapraro ❗️

Issue #23, opened 2018-11-17
Additional Thoughts on the Introductory Article


guywmartin:
#22 

Here are some additional thoughts on this first introductory article:

Overall, I think this is a great start - it's thorough and covers the topic in an appropriate level of detail.  

Some potential issues I found were:

- Should this be an article in the 1st person (who is I or we?)

- On Line 214, a statement was made about InnerSource being done during working hours, vs. 'after hours' contributions for Open Source.  That may be the case in some companies, but there are a lot of companies where InnerSource and Open Source contributions happen during working hours.  The vast majority of open source software is built by engineers working on company time to achieve business goals.  I think we should consider removing or rewording that section.

- There are a couple of descriptions of 'Newbies' or 'Newbie tasks', which I think sound a bit demotivating to those who may be 'newbies'.  Let's brainstorm a new name for those community members.

-  There is a lot of focus on TCs being community managers, as well as deep technical experts.  In some communities, you may be able to consider separating the roles so that a dedicated community manager can take on some of those.  This could be especially important if a TC has a deep technical passion, for example, but has a hard time interacting or promoting the project within an internal organization.

gruetter:
/CC @guywmartin , @rrrutledge , @cewilliams , @maxcapraro 

> Should this be an article in the 1st person (who is I or we?)

My gut feeling is to go with _we_ (the InnerSourceCommons Community) instead of _I_. Ok for everybody?  I used _I_ because the article served as the script for the TC video recording.

> [...] InnerSource being done during working hours, vs. 'after hours' contributions for Open Source [...]

You are right, @guywmartin. I propose to remove this. Any objections?

> [...] Newbie tasks [...]

I wasn't aware that _Newbie_ would be demotivating. Some ideas (from a non-native speaker).
Instead of _Newbie_:
- _Beginner_
- _Novice_
- _Newcomer_ (my favorite)

Instead of a _Newbie task_ tag:
- _Good for beginners_
- _First contribution_

Other ideas, everybody?

> There is a lot of focus on TCs being community managers, as well as deep technical experts. In some communities, you may be able to consider separating the roles so that a dedicated community manager can take on some of those. This could be especially important if a TC has a deep technical passion, for example, but has a hard time interacting or promoting the project within an internal organization.

[...]

PR #26, opened 2018-12-12
Incorporated feedback from #23

gruetter:
Thanks to @guywmartin and @rrrutledge for your feedback on the TC article. I have incorporated almost all of it, but there are still two sections which contain my personal experience and some personal examples. I'll highlight those sections with a remark. We need to decide whether or not we want to keep them (I'm fine with removing them, as they are in the video).


gruetter:
@rrrutledge @guywmartin : If you have time, could you pls. respond to the two open questions above?

rrrutledge:
Sounds great, I'm just not clear which two questions you mean?

gruetter:
I was referring to the question whether or not to remove the personal examples from the article altogether. I realise I've already done that with 246bfc8, though. So we're good to merge, IMO, @rrrutledge .

rrrutledge:
Oh, I accidentally committed some updates directly to the branch.  I meant to commit to a fork so that I could make a PR to the branch under review here.  Hopefully those edits look OK.  If not we can revert.

trusted-committer-article.md
```diff
@@ -490,21 +465,22 @@ contribution to the community and accelerate your own growth.
...
-the importance of having capable TCs for the success of any InnerSource 
-initiative and the level of empowerment that this role requires.
+to making contributions, upleveling the community and advocating the
+community's needs in their organization. We also talked about how to become a
+TC and what it takes to fill that role. Working as a TC will be demanding but
+it will also be very fulfilling and will help you amplify your value contribution
+in your company.
+
+In that sense, we hope that this video inspired you to set off on a path
+towards becoming a TC as well. And if you're doubt, why not just try it out?
```
rrrutledge: > And if you're doubt, why not just try it out?

I think that this line works out well in the spoken text, but feels a little off-tone here in writing.  Recommend just removing it.

gruetter:
> I think that this line works out well in the spoken text, but feels a little off-tone here in writing. Recommend just removing it.

Agreed, @rrrutledge . What do you think: should we also remove the reference to the _video_ in line 474?

rrrutledge:
> should we also remove the reference to the video

Agree.  For expediency just went and made the changes.  Feel free to revert or if it looks good to you and @guywmartin we can merge.

gruetter:
Ok great, @rrrutledge and @guywmartin . Thanks a lot for the review.

PR #32, opened 2019-01-11
Final update to slides shown in TC video

gruetter:
I've replaced the play button icon in the introduction segment with a film strip. That way it doesn't look clickable anymore.


rrrutledge:
Looks good to me!

Issue #33, opened 2019-01-16
Content Outline


rrrutledge:
Per [the schedule](https://docs.google.com/document/d/1QtjQ-qy9bBaPob3V1OldjENTqvj6USIQCjMomKs0XEQ) we'd need an outline for the content of the **Contributor** section of the learning path.

The outline should include a listing of the segments of the learning path as well as the major points of each segment.

Once this outline is finished we can simultaneously start producing both the written and video portions of the segments.

rrrutledge:
Started brainstorming some ideas in [these notes](https://paypalflow.slack.com/archives/CARTU4XV2/p1548443862048400).

KathrynKLewis:
The outline overall looks very good and comprehensive. My only suggestion is to add further detail for clarity to the section: 
- Understand that you have to ‘sell’ your idea - communicate early and often -before- investing a lot of time in coding an idea (fail fast)

What best practices might be shared so that the reader/contributor knows how to "sell" their idea? I imagine there is a framework or protocol to follow so that the contributor is helpful to the owner rather then needy/burdensome.

rrrutledge:
Thank you, @KathrynKLewis

guywmartin:
Thank you @KathrynKLewis - I'll add a quick note that we need to address this in the final video script.

PR #37, opened 2019-02-07
Create contributor/outline.md

lenucksi:
Sources:
 - as per the Google Doc by @guywmartin
 - distilled from the brainstorming notes linked in #33 

Please add your reviews, additions and discussion here.

### DIFF
--- a/contributor/outline.md
+++ b/contributor/outline.md
@@ -1,0 +1,46 @@
+# InnerSource Contributor Content Outline
+
+## Introduction/Rationale
+   - Quick Overview of Roles
+      - Explain how contributor role relates to other roles
+   - Discuss the value of contribution
+      - ‘With many eyes, all bugs are shallow’
+      - Improve cross-company collaboration
+      - Reduce single points of failure (team or personal)
+   - Contribution Rationale
+      - Prevent/reduce duplication of efforts through many teams
+      - Increasing personal reputation
+      - Things that you gain as a contributor.
+         - Help from others on the project in implementing your feature.
+         - Help from the broader community in maintaining what is written.  You are not stuck with the long-term burden of maintenance.
+         - Heavy influence into the final product.
+         - Assurance that the final product will be useful for you.
+
+## Mental/Attitude/Communications
+   - A lot of times I want to make a contribution but it’s hard to be OK with them saying “no” or that their goals are different then mine as a contributor.  How do you walk away from that feeling good.
+      - Adjust and have realistic expectations
+      - Empathy and humility
+   - Have a contributing mindset - think outside your project and look for places to contribute instead of reinventing the wheel.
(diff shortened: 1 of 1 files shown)


contributor/outline.md
```diff
@@ -0,0 +1,46 @@
+# InnerSource Contributor Content Outline
+
+## Introduction/Rationale
+   - Quick Overview of Roles
+      - Explain how contributor role relates to other roles
+   - Discuss the value of contribution
```
mrsanz: Seems like we need to also consider the value it brings to the person contributing as well.  I'd like to propose we add the overtly obvious to the list.  

- Directly increasing the impact for the contributor(individual).
- Through the contribution and review process, a contributor can and will level up their own skills, under the mentorship of the Trusted Committers.

contributor/outline.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Mental/Attitude/Communications
+   - A lot of times I want to make a contribution but it’s hard to be OK with them saying “no” or that their goals are different then mine as a contributor.  How do you walk away from that feeling good.
+      - Adjust and have realistic expectations
+      - Empathy and humility
+   - Have a contributing mindset - think outside your project and look for places to contribute instead of reinventing the wheel.
+      - A contributing mindset starts with a shared use mindset.  You’ll likely have to be comfortable using other folks’ stuff before it is realistic to contribute to it.
+         - Think about contributing.  Every time you’re working on something in a project that is not related to the purpose or domain of that project, look for the proper home for that code.
+   - Use the Guest in a home analogy.
+   - Nothing too small to contribute!!!!!!
```
mrsanz: might be important to break this apart and state what is a contribution.  Often times things aren't viewed as a contribution such as an idea, some code review, or documentation.

contributor/outline.md
```diff
@@ -0,0 +1,46 @@
...
+   - Understand that you have to ‘sell’ your idea - communicate early and often -before- investing a lot of time in coding an idea (fail fast)
+   - ‘Selling’ your participation to your management team
+   - Understand the general purpose of the project enough to know what your contribution makes sense as an overall part of the project instead of just your personal snowflake
+
+## Technical Concerns
```
mrsanz: I'm not sure what separates this advice from the other sections.  How are these things technical?

Perhaps rename this section to something like
> Working with code

rrrutledge:
Made some suggestions directly in the doc..

contributor/outline.md
guywmartin: I concur with this.

contributor/outline.md
guywmartin: Concur with this as well - I thought we had a section on contributions being more than code though?

contributor/outline.md
guywmartin: Agreed that it might not be the best place for them.  But, per your earlier point, I don't think we want to have a 'working with code' section if we want to allow for contributions that are broad (docs, bug reports, questions answered, etc.).

Issue #12, opened 2018-10-15
Written outline


rrrutledge:

rrrutledge:
@ttyler01 interested in taking a crack (backed up by @mrsanz)

maxcapraro:
I renamed this to `video outline` to avoid confusion (initially thought it is a duplicate of #27 ). Please feel free to rename if I misunderstood something.

maxcapraro:
Renamed it back as #10 appears to be the video outline for "trusted commiter". Sorry.

rrrutledge:
The idea was to produce an outline of the written content for quick review before taking the time for a  full-blown article.  Which I think we did at some point.

lenucksi:
It's in done, thus can be closed.

PR #38, opened 2019-02-11
Moved trusted committer files to the proper folder structure

lenucksi:
## Rationale
Here comes an attempt to start the following tasks:. 
- Move the trusted committer parts into their own proper folder structure
- Create an outline.md as might be referenced by #12.
- Replace all mentions of "InnerSource" with "inner source" - as discussed on Slack (if I find the matching messages, I'll link them here) also with @gruetter. 

## Caveats
- Maybe we should first try to merge in #22 and figure out the differences with the current master
- I'm happy to close this PR and separate this into multiple PRs as we might have different topics mixed into one PR here.


lenucksi:
As discussed on the last learning path call (see Google Doc), I've separated out the question of space or no space. #1 still has an outline for the trusted committer. I'd either move it with that PR or merge that PR and then move the outline together with this one.

rrrutledge:
What's here is good.  Just because there's more good that could be added should not stop this item from moving forward.

rrrutledge:
Merged.

PR #52, opened 2019-03-07
Dangling notes document around hosts&TCs and project documents.

lenucksi:
I found a bunch of non-contributor centered ideas / points that I had on the contributor brain storm document that might still be worth to be seen and added somewhere (in a later revision / update of text parts)

Most of that is meant for discussion and centered around 
 - general project approaches
 - project documents such as CONTRIBUTING.md and README.md.

It might actually be a good idea to highlight the value and importance of quality documents there and provide a bit of guidance how to create them. I'm happy to discuss my experiences around those documents.


rrrutledge:
Hi, @lenucksi, as always your thoughts are excellent, sir.  As wonderful as the content is, here, we don't really have any kind of mechanism to take these ideas forward if we put them here.  What I'd recommend here is redirecting these thoughts to another location.  The two things I can think of are:

* If the content would fit in pretty closely with one of the Trusted Committer videos, then PR it to the associated article.
* Otherwise, let's see if it can fit into one of our InnerSourcePatterns (or create a new pattern).

This learning path won't be able to house everything we know and keep learning about inner source, but the _InnerSourcePatterns_ repo by rights should grow to be the root of all knowledge on inner source.

lenucksi:
> Hi, @lenucksi, as always your thoughts are excellent, sir. 
Thanks!
> As wonderful as the content is, here, we don't really have any kind of mechanism to take these ideas forward if we put them here.

I figured the same but wanted to put them up for discussion of validity at some point and GitHub sounded like a valid option. 

 What I'd recommend here is redirecting these thoughts to another location. The two things I can think of are:
> 
> * If the content would fit in pretty closely with one of the Trusted Committer videos, then PR it to the associated article.

I'll see if I can fit some of the content somewhere and add a PR if possible.

On the case of the documents:
I have seen multiple cases where people were not sure what to best put in such a document and were grateful for suggestions. I'd suggest to add pointers to places that have such information so people can inform themselves about. I've got a small set of templates maybe I can provide them as a link target, will file a PR for the trusted committer text then.

> * Otherwise, let's see if it can fit into one of our InnerSourcePatterns (or create a new pattern).
> 
> This learning path won't be able to house everything we know and keep learning about inner source, but the _InnerSourcePatterns_ repo by rights should grow to be the root of all knowledge on inner source.

True and good idea, I'll see what I can integrate there.
Maybe we find a bunch more options to point to the patterns repo in the learning path so people instantly have it on their mind as a go-to location after doing the learning path.

rrrutledge:
> GitHub sounded like a valid option

That is true.  I could see discussing them here via this PR and then when thoughts are refined we see where they can go?

 > On the case of the documents

I don't understand - to which documents are you referring?

> Maybe we find a bunch more options to point to the patterns repo

This would be great!  In general I feel that we need more refined and polished ideas in the _InnerSourcePatterns_ repo.  I wish I had more time to work there, but hope to have that time freed up after finishing the learning path.

lenucksi:
> > GitHub sounded like a valid option
> 
> That is true. I could see discussing them here via this PR and then when thoughts are refined we see where they can go?

That is pretty much what I intended! :+1:  

> 
> > On the case of the documents
> 
> I don't understand - to which documents are you referring?

CONTRIBUTING.md and README.md in that case. We refer to them (at least the former) in various places. I've seen that not everyone knows how create good, useful versions of them and people happy about having a small "template" asking some questions to start from.

> 
> > Maybe we find a bunch more options to point to the patterns repo
> 
> This would be great! In general I feel that we need more refined and polished ideas in the _InnerSourcePatterns_ repo. I wish I had more time to work there, but hope to have that time freed up after finishing the learning path.

:+1: 
I think the pattern format acutally can be or is a pretty nice way to couple the academic research and practical application perspectives.

rrrutledge:
> CONTRIBUTING.md and README.md in that case

Do we feel that there would be much difference in what we recommend in inner source compared to what is recommended for open source?  If not then I think that we should find some existing documentation and point to that.  https://todogroup.org/ is a good place to start for finding such existing documentation.

lenucksi:
> > CONTRIBUTING.md and README.md in that case
> 
> Do we feel that there would be much difference in what we recommend in inner source compared to what is recommended for open source? If not then I think that we should find some existing documentation and point to that. https://todogroup.org/ is a good place to start for finding such existing documentation.

I would be entirely happy to point to other resources too, some pointers should be there though. TODOGroup might be entirely fine.

From what I have seen, some emphasis on expectation management with regards to reaction times might be warranted as some InnerSource projects in globally distributed companies having to deal with e.g. US-West-Europe or even US-West-India time differences sometimes experienced more tensions due to implicit infeasible reaction times.

We also have https://github.com/paypal/InnerSourceCommons/wiki/CreatingAContribMd and https://github.com/paypal/InnerSourceCommons/wiki/CreatingAReadme sitting around which might be a bit sparse but catches quite some good points.

rrrutledge:
Here's the closest thing I could find: https://opensource.guide/starting-a-project/.  I agree that such a document needs to exist and it could be hosted in _InnerSourceCommons_ if it doesn't exist anywhere else.

lenucksi:
> Here's the closest thing I could find: https://opensource.guide/starting-a-project/. I agree that such a document needs to exist and it could be hosted in _InnerSourceCommons_ if it doesn't exist anywhere else.

The link has good content, and I think we should provide it in a refence section. We definitely need one. Either at the end of the complete learning path (literature section style) or as footnotes / per individual module of the learning path. 

I have a document "template" (basically a file with some structure suggestions asking a few questions) that I synthesized from a number of external documents (I think some of the documents in the link above were actually included in the research back then) and am using internally with some success. Maybe I can contribute that here.

rrrutledge:
Sounds very useful.  I think it would be good to have a conversation in [#innersource-patterns](https://paypalflow.slack.com/messages/C2EFRTS6A/) about where to put it.  It's not exactly a pattern - more like a how-to guide, but still needed and useful.

rrrutledge:
[README](https://patterns.innersourcecommons.org/appendix/extras/readme-template) and [CONTRIBUTING](https://patterns.innersourcecommons.org/appendix/extras/contributing-template) guides exist in _InnerSource_ patterns.  Seems that any feedback from this PR can go there?

PR #55, opened 2019-03-13
Split Trusted Committer articles

rrrutledge:
Allows us to link to each one separately.


lenucksi:
This looks like a good idea, thanks for doing that. I'll need to see when I can look into the details, hopefully today.

rrrutledge:
Thanks, @lenucksi.  It should just be a straight copy from one location to the next.

lenucksi:
I came around today, this is as you said a clean separation of the one big file. One suggestion though: How about putting a number in front of the articles, e.g. ```01-introduction.md``` instead of ```introduction.md``` such that the order that was implicitly given in the original file is preserved?

rrrutledge:
> How about putting a number in front of the articles

Good idea.  Done.

rrrutledge:
> [lenucksi](/lenucksi) merged commit [`45860f4`](/InnerSourceCommons/InnerSourceLearningPath/commit/45860f4038e08fe5d3a98c9be91c07e7411b1a95) into master

Thanks!

PR #63, opened 2019-03-25
Document meeting roles

rrrutledge:
* [This document](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/meeting-roles/meetings/roles.md) describes everything that I am doing to run the meetings of the Inner Source Learning Path working group.
* The document partitions the activities involved in running a successful meeting into a set of loosely-coupled, highly cohesive roles.
* While I am happy to continue to fill all of these roles, I feel that as the working group matures it will make sense to share leadership with other members of the community via filling these meeting roles.

### DIFF
--- a/meetings/roles.md
+++ b/meetings/roles.md
@@ -13,19 +13,19 @@ The roles are:
 The _Scheduler_ sends the calendar invite for the meeting.
 
 **Within 24 Hours of Previous Meeting**
-* [ ] Create an agenda document for the next meeting by copying [template][template] into the [learning path folder].
-* [ ] Open the document and follow the steps in _Do these things to set up this document_.
+* [ ] Create an agenda document for the next meeting by copying this [template][template] into the [learning path folder].
+* [ ] Open the document and follow the steps in the "_Do these things to set up this document_" section.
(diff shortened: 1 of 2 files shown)


meetings/roles.md
```diff
@@ -0,0 +1,138 @@
...
+This invite must include:
+
+  * Conference call info.
+  * Comment-enabled link to the agenda document.
+
+### Tips
+
+* The _Scribe_ should have included the time of the meeting in the notes from the previous meeting. 
+* Include everyone from the previous meeting on the invite for the next meeting.
+An easy way to do this using _Outlook_ is to reply by meeting to the notes email from the previous meeting.
```
lenucksi: There are other PIM programs, and likely most of them contain the _"Reply with meeting"_ button and can display the initial email (At least Evolution and Outlook do) - thus I would phrase this more generally 
```suggestion
An easy way to do this is to use the _Reply with meeting_ function to the notes email from the previous meeting.
```

rrrutledge:
Thanks for this review, @lenucksi.  Since other reviews might come in before this week's meeting I'll wait until that time to merge.

PR #80, opened 2019-05-06
Update Scribe

rrrutledge:
* Encourage group writing of notes and assignments.
* In the end the **Scribe** is still responsible for ensuring that the notes are there and good, regardless of who has written them.


lenucksi:
That looks good to me.

PR #83, opened 2019-05-14
Send out write-enabled link

rrrutledge:
* Feedback from 5/9 call.
* Also curious for review from @mtabbal


rrrutledge:
Awesome!

PR #85, opened 2019-05-16
Rewording

rrrutledge:
Typo and breaking up long sentence.


rrrutledge:
> Isn't the first one overwriting the planned changes

Not intended.  I was just trying to break up a long sentence into bulleted form.  If any language was inadvertently changed then we should bring it back in line with the original.

lenucksi:
> > Isn't the first one overwriting the planned changes
> 
> Not intended. I was just trying to break up a long sentence into bulleted form. If any language was inadvertently changed then we should bring it back in line with the original.

Will manually fix this up then and get it committed.

rrrutledge:
Thanks, Johannes.

PR #86, opened 2019-05-24
Updating for recurring meeting invite

rrrutledge:
Scheduler no longer needs to create a brand-new meeting invite.

### DIFF
--- a/meetings/roles.md
+++ b/meetings/roles.md
@@ -18,7 +18,7 @@ The _Scheduler_ sends the calendar invite for the meeting.
 * [ ] Select `Make a copy`.
 * [ ] Replace `MM` and `DD` in the document name with the two-digit month and day of the upcoming meeting.
 * [ ] Open the document and follow the steps in the **Do these things to set up this document** section.
-* [ ] Send the calendar invite for the next meeting.
+* [ ] Update the calendar invite for the next meeting.
 This invite must include:
 
   * Conference call info.
@@ -26,10 +26,6 @@ This invite must include:
 
 ### Tips
 
-* The _Scribe_ should have included the time of the meeting in the notes from the previous meeting. 
-* Include everyone from the previous meeting on the invite for the next meeting.
-An easy way to do this is to use the _Reply with meeting_ function to the notes email from the previous meeting.
-* The title of the meeting should be "Inner Source Learning Path Working Session".
 * You can use your favorite corporate conference call system or create a Google Hangout.
 * Check out this [writeup][gdoc sharing] if you need help creating a write-enabled link to the agenda document.
 


meetings/roles.md
```diff
@@ -18,18 +18,14 @@ The _Scheduler_ sends the calendar invite for the meeting.
...
+* [ ] Update the calendar invite for the next meeting.
 This invite must include:
 
   * Conference call info.
   * Write-enabled link to the agenda document.
 
 ### Tips
 
-* The _Scribe_ should have included the time of the meeting in the notes from the previous meeting. 
-* Include everyone from the previous meeting on the invite for the next meeting.
```
lenucksi: Actually I think these are still good tips. 
We are still updating the invite anyway (as I understood it) with
- a new Google-Doc-Link (that could be prevented if we were to have a innersourcecommons.org-redirect that always points to the right Google Doc, although that would just add another layer of indirection. Praise the RFC1925 btw ;).)
- possibly any new participants / interested people / ... recipients addresses 
which is what the tips describe. Thus I think the tips maybe need to be rephrased but should be kept otherwise.

rrrutledge: Yes the _Scheduler_ needs to update the _body_ of the invite but not its time or attendees.

We haven't documented how attendees get on the recurring invite - probably should be something the _Scribe_ does after each meeting as part of sending out the notes.  Add anyone new that showed up to the recurring invite.

lenucksi: Valid points. They bring me to the point of asking how we do a distributed update of the invite (or parts of it). Do we want to use a shared calendar with the invite and permissions on it? Can one just reply differently and update the series?

rrrutledge: Yes, I haven't written this down, either, but I was thinking of creating a shared Google account that we all can sign into so that anyone can make updates to an instance of the calendar series.

lenucksi: I'm sure that Google would support shared access to a calendar such that we would not need to actually share a set of credentials...
The changes are fine with me then. Do we want to address the modified _Scribe_ role (as per this discussion) in this PR too (and thus not merge until the details are in)?

rrrutledge: > Do we want to address the modified Scribe role 
I'd suggest waiting since this depends on creating some kind of shared mailbox, which hasn't been done yet.

lenucksi: Based on reading this discussion, it appears to me that:
- A shared calendar with r/w access for the Scribe and the Scheduler
- No shared mailbox as mails can be send from whomever to the list and the recipients based on the last mail.

The internet had the following on doing that with a Google product:
- Calendars: Can be shared with other Google account holders, even with the free version. It is possible to delegate almost full r/w access. That could be sufficient for us.
 - https://support.google.com/calendar/answer/37082
- Mail: Delegation is only possible with GSuite products. There is also Google Groups for business
 - https://support.google.com/mail/answer/138350
 - https://support.google.com/a/answer/167430

(Also this discussion should probably continue in an issue...)

rrrutledge:
Thanks, @lenucksi!.  Created [Shared Calendar for Learning Path invites](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/89).

Issue #39, opened 2019-02-12
Decide upon a spelling of 'InnerSource' vs. 'inner source' within the articles and clean up accordingly.


lenucksi:
As mentioned in PR #38 and discussed on [Slack](https://paypalflow.slack.com/archives/CARTU4XV2/p1549613024083900?thread_ts=1549581428.081200&cid=CARTU4XV2) we need to decide which spelling to use - either 'InnerSource' or 'inner source'.
After deciding on that we'll need to clean up the articles respectively. 

Let's use this ticket to coordinate / bundle all related matters, i.e. close this
- once the decision is made **AND**
- all files are cleaned up accordingly.

rrrutledge:
Thanks for opening, @lenucksi

rrrutledge:
In the last InnerSource Commons Governance meeting we said that we would try out using InnerSource (1 word) in its noun form only.  When a verb is needed we could go through and pair another verb with the noun InnerSource (e.g. "adopting InnerSource" instead of "InnerSourcing").

fioddor:
For reference:
- https://www.youtube.com/watch?v=6nFNnDKTQlg
- https://github.com/jeffabailey/InnerSourcePatterns/blob/main/meta/innersource-spelling.md

rrrutledge:
Thank you!

just curious … how did you find this?

PR #91, opened 2019-06-11
Addressed #39's decision for the introduction articles

lenucksi:
Modified the introduction articles according to #39's decision to use InnerSource instead of 'inner source' where possible. 
I had to modify the sentence in a few cases but many other cases worked in a search&replace for my sense of writing. 

@rrrutledge, @arnom-ms and @NewMexicoKid : I'd be happy if you could provide me with your take on the language and see if I have overseen any instance of the word.


lenucksi:
This got superseded by #91

PR #92, opened 2019-06-13
* Trivial fixing of typos etc

arnom-ms:
Almost all occurrences of "inner source" were already converted to "InnerSource"


lenucksi:
Looks good to me, thanks for the typo fixes, too. Also makes #93 superfluous as this PR also contains the introduction article adaptations.

Issue #16, opened 2018-10-19
Post Production


rrrutledge:

rrrutledge:
@Silona interested in authoring

rrrutledge:
@maxcapraro as dedicated reviewer.

rrrutledge:
Here is the preview of the `Product Owner` videos: https://oreillymedia.screenlight.tv/shares/yrAjQsGWh8HAOWOZhssw2BP9KXdU6Tgt?_=1541987258945=

rrrutledge:
As the video has been shot already and is "done", feedback should focus on the slides and graphics and if those make sense.  Updates can be made there if needed.

maxcapraro:
I added my feedback to this file: 
[Product Owner Video Feedback.txt](https://github.com/InnerSourceCommons/InnerSourceLearningPath/files/2640825/Product.Owner.Video.Feedback.txt)

In general: Not much changes are needed. Excellent job @silona and all involved :)

rrrutledge:
Thanks for this review, @maxcapraro!  I will have a look as well!

rrrutledge:
## All videos
* Suggest finishing with the InnerSource Commons logo and URL (similar to the "Introduction" videos)
* Suggest using flyovers instead of cutting to a slide deck when you just need a few works to show.  Then we can keep seeing the face of the presenter.
Similar to "Introduction" videos.

## 02_primer
04:00 -Use the bit.ly short link to refer to the learning path - http://bit.ly/inner-source-learning-path

## Video 04_body_p2
04:14 -  Capitalization of “Innersource” is not consistent.  Should be "Inner Source"

## Video 06_recap_takeaways
48:00 - Should be plural, not possessive – TCs, POs.

## Video 07_contact
04:00 - Use the inner source logo in the flyover text (similar to "Introduction" segments).
13:00 - Use the bit.ly short link (http://bit.ly/inner-source-learning-path).

maxcapraro:
Our friends at O'Reilly have provided the updated videos:

- 01_Opening_revised
- 03_How Difficult_revised
- 04_Major Benefits_revised
- 05_New Roles_revised
- 06_Recap_revised

I reviewed them and they look good to me :+1:

rrrutledge:
Thanks, @maxcapraro !

lenucksi:
This is indeed very solid content. I agree with @maxcapraro: Very nice job @Silona and all involved people! :+1: 

[...]

PR #93, opened 2019-06-25
Post Production - Issue 16

stuartpurgavie:
https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/16
Update to use "program" instead of "project" to align with voice track
On slide titled "Top 5 Common Complaints"
Update in master presentation document and exported screen captures


lenucksi:
> #16
> Update to use "program" instead of "project" to align with voice track

I see it updated as ```product``` instead of ```project``` - it sounds plausible and I haven't heard the audio for a while, but is that desired?

stuartpurgavie:
@lenucksi the purpose of this PR was to bring the audio and visual tracks in line. I'm not able to modify the voice track using a PR, so this was all I could do to promote consistency.

If word choice is a concern of yours, then we could define the words and then get specific? From my understanding of IT PMO, "program" is an overarching term to describe a collection of "projects" that implement or modify an application stack for a given purpose, so the use of "program" here may be appropriate. However, it could also be a jargon use that is more confusing for technical audience members, as for many "program" and "application" are analogous terms.

lenucksi:
> @lenucksi the purpose of this PR was to bring the audio and visual tracks in line. I'm not able to modify the voice track using a PR, so this was all I could do to promote consistency.

I'm fully onboard with this and think it's good. 

My comment should've read
> I see it updated as ```product``` instead of ```program``` - it sounds plausible and I haven't heard the audio for a while, but is that desired?

As in, your original PR description said "update to program" but the new slide says "product" which is incosistent and the reason for me to ask I something at either location was a typo. Maybe I also just didn't find the place where it was changed as described in the PR. 

> If word choice is a concern of yours, then we could define the words and then get specific? From my understanding of IT PMO, "program" is an overarching term to describe a collection of "projects" that implement or modify an application stack for a given purpose, so the use of "program" here may be appropriate. However, it could also be a jargon use that is more confusing for technical audience members, as for many "program" and "application" are analogous terms.

I agree with you that this is a potentially confusing overlap of words but not something I would change in the slides or address here. Maybe in a sentence in the article that goes with the video.

lenucksi:
#96 is relevant for the assignments.md file here.

rrrutledge:
Sorry folks, I've been on vacation, but I just double-checked and the slides in this pull request now match the current audio.

rrrutledge:
Great work, @stuartpurgavie !

PR #95, opened 2019-07-04
Update the roles document for shared calendar use.

lenucksi:
Added a few changes to reflect the use of the the shared calendar to coordinate the learning path calls.


meetings/roles.md
```diff
@@ -112,7 +112,7 @@ This email should include (in order):
   * Link to the agenda document where the notes are located.
   * Chatham House Rule disclaimer.
   * Copy/pasted notes from the agenda document.
-  
+* [ ] Possibly update the invite with new participants emails.
```
rrrutledge: This point is a good one.  I feel like this could be the *Scheduler*, though.  Before updating the next invite make sure that the series has everyone on it.

lenucksi: Fair point, I agree. The Scheduler is able to see who has been in the call from the document and should then be able to retrieve an email address to add to the invite.

Issue #67, opened 2019-04-30
Article: The Contributor in InnerSource


rrrutledge:
We should produce a written article based on **The Contributor in InnerSource** section of the [_Contributor_ outline](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/outline.md).

The article should teach the same points as [The Contributor in InnerSource Video](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/68).

lenucksi:
Move to done, thus closing the ticket.

PR #98, opened 2019-07-24
First draft for the Introductory Contributor Section Article #67

lenucksi:
### Content
This is an initial draft for the "The Committer in InnerSource" article based upon the outline and video proposal document.
Feedback, Change proposals either using the GitHub change feature or as pushes to the branch for larger changes (hence the direct push to this repo) are more than welcome.

### Technical detail
- I've included headers from the outline document and video proposal for drafting purposes. Once they document is ready, those should obviously disappear.
- The referenced articles currently have no URL yet, thus the link fields are currently left blank. They need to be replaced once they are ready. An initial approach could be to link the GitHub articles.

### DIFF
--- a/contributor/01-introduction-article.md
+++ b/contributor/01-introduction-article.md
@@ -1,0 +1,97 @@
+# Draft comments
+
+ Please watch the [Intro to InnerSource Video](https://www.safaribooksonline.com/videos/introduction-to-innersource/9781492041504)
+* Also watch the [Introduction](https://learning.oreilly.com/videos/the-trusted-committer/9781492047599/9781492047599-video323925) to the Trusted Committer role.
+* The [Lowering the Barriers to Entry](https://learning.oreilly.com/videos/the-trusted-committer/9781492047599/9781492047599-video323929) video of the Trusted Committer role can be a good addition to understand more of why the Trusted Committer might ask for different things.
+
+## Outline based Target
+### The Contributor in InnerSource
+
+  - Introduce the speaker. - only required for the video script I'd say. 
+  - Quick overview of the _Contributor_ role.
+  - Summary of the roles of InnerSource and how each relates to _Contributor_.
+  - Summary of the upcoming segments and what each covers.
+
+Outline for this section inspired by the [Trusted Committer introduction article.
+
+## Video Play-book based Target
+How do I best contribute to an InnerSource project?  This segment shares a summary of what it means to be an InnerSource Contributor.  We'll also give an overview of the upcoming training segments on being a great InnerSource Contributor.
+
+
+
+# Article
+
+## The Contributor in InnerSource.
+
+Without contributions, an InnerSource community and its project(s) would not exist in a meaningful way.
+There would be open repositories, ideally equipped with somewhat specific extra documents (e.g. a CONTRIBUTING.md and a README.md) and person(s) committing to them, e.g. a Trusted Committer.
+[EDITOR NOTE:
+- In other articles we draw a distinction between guest and host team.
+- We argue that the Committer is part of the guest team and needs to fit in with the host team.
+- In a more-than-one-person project what do we call the people aside of the TC/PO that are members of the host team? Just team members?
+        - OSS would likely not make this distinction and everyone would end up as contributor until it's committer / trusted committer at some point I guess?
+        - In the ASF-Rule-set this would be the distinction between Developer and Committer.
+]
+
+The Contributor role describes a person that makes contributions to an InnerSource community.
+A contribution can be a code or non-code artifact such as a bug-report, an addition to documentation or a feature request. 
+[EDITOR-NOTE: IMPORTANT we should probably distinguish that just requesting is not already InnerSource or phrase this accordingly].
+
+This person may or may not be part or see themselves as part of the community.
+However, for quite a few people there is a sort of journey contributors can make from just knowing about the community to using the community's product to interacting with members of the community and finally starting to contribute.
+Finally, some of them may even become Trusted Committers.
+
+
+## Summary of the roles of InnerSource and how each relates to _Contributor_.
+
+As a Contributor in an InnerSource community you will interact with people playing other roles of InnerSource, such as Trusted Committer or Product Owner and possibly also other contributors.
+At times, these roles can be played by the same person, such as Trusted Committer and Product Owner in small grass-roots style projects.
+
+This section gives you a very short overview of the other two roles but we'd like to encourage you to read the [introductory article of the Trusted Committer role]() and recommended to read the [Lowering the barriers to entry]() article too before you delve deeper into the details of the Contributor role in this section.
+You can also watch the videos ([introduction](intro-trusted-committer),[lowering the barriers to entry](tc-lowering-the-barriers-to-entry)) instead of the articles.
+
+### Trusted Committer
+
+A Trusted Committer will be your host for your stay in the hosting community.
+They are the gatekeepers to the project's code repository and will move your contribution closer to production once they accepted it. 
+It is their role to mentor you on your way to contributing to their community, e.g. by providing good information to enable you to help yourself or help you in other cases.
+Such information could be established house rules for reviews, proposal of larger changes,  pointers to documentation or code sections relevant for your contribution.
+
+They also need to care about product quality, sustainability and project evolution from a technical as well as general perspective, about reducing the barrier to making to contributions for everyone as well as about caring for their community in general.
+Caring for the community involves keeping it healthy, upleveling it and its participants and advocating its needs in their organization.
+
+
+### Product Owner
+[EDITOR NOTE: We actually have no defined articles here. Below rough draft is based upon https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/96/files#diff-1fc7c3a09a5ec014a5ba5313dc6473fa]
+The role of the Product Owner has some similarity with the product owner role of your average project.
+However, there are some differences - depending on the size of the project, this role is often filled by the same person that acts as trusted committer.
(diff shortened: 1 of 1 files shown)


contributor/01-introduction-article.md
```diff
@@ -0,0 +1,97 @@
...
+## The Contributor in InnerSource.
+
+Without contributions, an InnerSource community and its project(s) would not exist in a meaningful way.
+There would be open repositories, ideally equipped with somewhat specific extra documents (e.g. a CONTRIBUTING.md and a README.md) and person(s) committing to them, e.g. a Trusted Committer.
+[EDITOR NOTE:
+- In other articles we draw a distinction between guest and host team.
+- We argue that the Committer is part of the guest team and needs to fit in with the host team.
+- In a more-than-one-person project what do we call the people aside of the TC/PO that are members of the host team? Just team members?
+        - OSS would likely not make this distinction and everyone would end up as contributor until it's committer / trusted committer at some point I guess?
+        - In the ASF-Rule-set this would be the distinction between Developer and Committer.
```
lenucksi: This distinction is taken from https://www.apache.org/foundation/how-it-works.html#developers

rrrutledge:
Thanks for this pull request, @lenucksi.  I think that I'll be able to have a look before this week's call for the learning path.

rrrutledge:
> * An initial approach could be to link the GitHub articles.

Yes we should do this

contributor/01-introduction-article.md
```diff
@@ -0,0 +1,97 @@
...
+# Article
+
+## The Contributor in InnerSource.
+
+Without contributions, an InnerSource community and its project(s) would not exist in a meaningful way.
+There would be open repositories, ideally equipped with somewhat specific extra documents (e.g. a CONTRIBUTING.md and a README.md) and person(s) committing to them, e.g. a Trusted Committer.
+[EDITOR NOTE:
+- In other articles we draw a distinction between guest and host team.
+- We argue that the Committer is part of the guest team and needs to fit in with the host team.
+- In a more-than-one-person project what do we call the people aside of the TC/PO that are members of the host team? Just team members?
```
rrrutledge: I think that something like _host team members_ is fine.  Is this concept used somewhere in this article?

contributor/01-introduction-article.md
```diff
@@ -0,0 +1,97 @@
...
+## Video Play-book based Target
+How do I best contribute to an InnerSource project?  This segment shares a summary of what it means to be an InnerSource Contributor.  We'll also give an overview of the upcoming training segments on being a great InnerSource Contributor.
+
+
+
+# Article
+
+## The Contributor in InnerSource.
+
+Without contributions, an InnerSource community and its project(s) would not exist in a meaningful way.
```
rrrutledge: This paragraph makes the case for why contributors are needed and useful, which is great.  I wonder if it would be good to change the perspective, though.  I feel like this perspective assumes that one is already committed to having a vibrant array of InnerSource communities at the company and wants to know how they can help that along.  There may be some people like that around, but I'd suggest that it's not the broadest way to appeal to potential contributors.

What about a more personal approach?  Something like:

```
Have you ever been blocked in your next coding task because another team didn't have time to add a feature in their system that you depend on?  Perhaps after a while you even had to do some extra work in your project to work around the missing feature.  How nice would it be to never be blocked in this way?

With projects that run via InnerSource, you'll never be blocked waiting for another team to deliver some needed feature.  If you're not getting what you need you can make the change you need directly in the other team's code repository by acting as an InnerSource Contributor.

...
```

rrrutledge: The exact words here that I'm suggesting aren't the main point as much as the persona to-whom we're talking and what we assume they value.  That's the real suggestion I'm trying to give, here.

contributor/01-introduction-article.md
```diff
@@ -0,0 +1,97 @@
...
+[EDITOR NOTE:
+- In other articles we draw a distinction between guest and host team.
+- We argue that the Committer is part of the guest team and needs to fit in with the host team.
+- In a more-than-one-person project what do we call the people aside of the TC/PO that are members of the host team? Just team members?
+        - OSS would likely not make this distinction and everyone would end up as contributor until it's committer / trusted committer at some point I guess?
+        - In the ASF-Rule-set this would be the distinction between Developer and Committer.
+]
+
+The Contributor role describes a person that makes contributions to an InnerSource community.
+A contribution can be a code or non-code artifact such as a bug-report, an addition to documentation or a feature request.
```
rrrutledge: I think that this introduction should frame a Contributor as someone that is making changes to the contents of a repo.  While true that opening bug or feature issues in the product backlog is a valuable addition, the whole thrust of this section, the learning path, and the InnerSource movement in general is to teach people have to make changes to the contents of a repo that they do not own.  This introductory article should phrase it that way.

[Becoming an InnerSource Contributor](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/cb22be30c19afac949a8bea988a14e805b4fd9c5/contributor/02-becoming-a-contributor.md) gives a small call-out of other ways to contribute, but I don't think we should highlight it up-front.

contributor/01-introduction-article.md
lenucksi: I agree with your point of the persona we're talking to - I guess this was a case of academic writing style coming through for a bit again. ;)
The [word suggestion](https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/98/files#r312260530) you made sounds fine to me.
Could you convert that to a commit suggestion? I'll merge and then we can continue from there.

contributor/01-introduction-article.md
```diff
@@ -0,0 +1,97 @@
...
+
+If you want to learn in more detail what these other roles are about, and we encourage you to do so, we've prepared sections about the [Trusted Committer](link) and the [Product Owner](link) too.
+
+
+## Overview of this section and its segments
+In the following 5 videos you will learn more in detail about the various aspects introduced before. 
+
+The second segment, i.e. the one after this segment, will detail _mindset and habit_ that create opportunities to _become an InnerSource Contributor_.
+
+In the third segment we'll look into the _ethos of the Contributor_ - i.e. aspects of _behavior_ that will lead to a pleasant and productive time for you and the home team and might spark more collaboration.
```
lenucksi: Do we use Home team? Shouldn't that be host team?

contributor/01-introduction-article.md
lenucksi: We're using that concept implicitly in lines 41,42 and explicitly in line 55.
Also, we're announcing to use the concept extensively in line 85. 

And since home/guest might establish a mental concept of separation (us/them) I think that we should give that a aspect a consideration to see if and how we want to deal with it throughout the content. 
This could enhance clarity and prevent contradictory information later on. (Me being a nitpick and definition pedant here ;))

I've suggested the [Apache source below](https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/98/files#r307215859) since that gives an example for this

contributor/01-introduction-article.md
lenucksi: I agree with your point - when bringing this later on we can leave it out in the introduction.
Could you make a commit suggestion from this?

rrrutledge:
> I'm happy for suggestions

Left in https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/110.

contributor/01-introduction-article.md
rrrutledge: https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/111

contributor/01-introduction-article.md
rrrutledge: Yes, you're right.

```suggestion
In the third segment we'll look into the _ethos of the Contributor_ - i.e. aspects of _behavior_ that will lead to a pleasant and productive time for you and the host team and might spark more collaboration.
```

contributor/01-introduction-article.md
rrrutledge: I think in a corporate context we assume that division is already built in by the organization.  We want to bring it together and suggest the terminology of _Host/Guest_ to encourage people to give warm welcome and collaboration to each other.

contributor/01-introduction-article.md
rrrutledge: Since it's multi-line made a new PR instead: https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/112

contributor/01-introduction-article.md
lenucksi: #112 fixed that as well.

contributor/01-introduction-article.md
lenucksi: Sounds good to me.
So we're using _host team member_ for a person that is part of the host team but no trusted committer (yet) then?

rrrutledge: Yes, I think so.

contributor/01-introduction-article.md
rrrutledge: Have a look at the updated PR.

contributor/01-introduction-article.md
rrrutledge: True

contributor/01-introduction-article.md
lenucksi: > Have a look at the updated PR.

Took a look, liked what I saw.

lenucksi:
> Looks good to me. We can revisit the `EDITOR NOTE` later.

Thanks, I'd say we put this up for merge discussion on the next call. 
I'd think the `EDITOR NOTE` was dealt with the term _host team member_ according to [this](https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/98#discussion_r313152374).
What do the others think?

(And we can still add new PRs on top of this later.)

rrrutledge:
@lenucksi we forgot on the call to discuss this article, specifically, if it can be merged.  If you think so then go ahead.

PR #100, opened 2019-07-25
Add the cost sharing aspect to company level benefits

MaineC:
Taken from Open Source projects: There are some where people collaborate because they would not be able to build the equivalent system in-house - so they collaborate to share the cost of development (or expressed another way to grow beyond what they could accomplish alone.)

### DIFF
--- a/contributor/outline.md
+++ b/contributor/outline.md
@@ -132,6 +132,7 @@ Outline for this section inspired by the [Trusted Committer introduction video].
     This typically makes the code more reusable, replaceable and secure.
     - Better alignment of shared projects to the needs of their internal consumers.
     - Improved and streamlined work environment retains more employees.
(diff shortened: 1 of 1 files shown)


contributor/outline.md
```diff
@@ -132,6 +132,7 @@ Outline for this section inspired by the [Trusted Committer introduction video].
     This typically makes the code more reusable, replaceable and secure.
```
rrrutledge: OK to include your point in the existing bullet on engineering efficiency?

```suggestion
    This typically makes the code more reusable, replaceable and secure.
    Collaboration across team boundaries saves duplicate effort and thus development cost
```

lenucksi: We did not have the "shared development cost" point explicitly stated here yet. However academic research and practical insight gives this as a selling point with direct business relevance. One that could be used by possible contributors selling the approach.
Engineering Efficiency feels similar but not equal and less explicit with that regard to me. My opinion: I would keep it as it's own point

PR #102, opened 2019-07-25
Fixes #72 adding detail to the contributor ethos.

### LINKED ISSUE #72: Article: The Contributor Ethos
We should produce a written article based on **The Contributor Ethos** section of the [_Contributor_ outline](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/outline.md).

The article should teach the same points as [The Contributor Ethos Video](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/71).

MaineC:
This follows the format of the introduction article, and the content suggestions from the outline.

### DIFF
--- a/contributor/03-contributor-ethos.md
+++ b/contributor/03-contributor-ethos.md
@@ -20,25 +20,70 @@ How do these expectations apply to InnerSource projects?
 
 #### Getting in
 
-In the real world there are several ways that guests are invited in:
+When visiting your neighbors you will likely not enter their home without
+knocking or ringing the door bell even if the door is open. Likewise in Inner
+Source as a visitor you won't be able (or invited) to commit directly to any
+code repository.
 
-* An "open" sign lets them know they can walk right in without knocking.
-* Knock on the door unexpected.
-* Knock on the door only when invited.
+Instead when visiting your neighbors you're likely to at least knock on their
+door. Likewise in Inner Source after making your changes to the codebase you'll
+submit them as a pull request. Much like you wouldn't go about making large
+changes and what you consider improvments to your neighbors home, in Inner
+Source you will anticipate and follow the projects collaboration guidelines. In
+turn your hosts will show you the way - in Inner Source that translates to
+existing trusted committers spending their time to mentor guests.
 
-In like manner, a contribution to an InnerSource project can have several ways that it gets accepted.
+What about those lovely summer parties that you went to? Likely before going
+there was some planning ahead of time to choose the right date and time, to
+prepare enough food or have that contributed by the guests. The same happens for
+bigger changes in Inner Source projects: Before making large modifications a
+project is likely to first ask you to submit an issue describing your need and
+your proposed solution. Spending some time on an intial design instead of
+jumping right into the implementation here saves contributors from having to
+re-do a lot of their work. Sharing progress early - even when it's not finished
+yet - helps mentoring the contributor towards a better solution. Much like
+[Yonik's law of unifinished
+patches](https://cwiki.apache.org/confluence/display/solr/HowToContribute)
+explains: "A half-baked patch in Jira, with no documentation, no tests
+and no backwards compatibility is better than no patch at all."
 
-Sometimes you just open a pull request. At times you may need to get a feature request accepted before opening the pull request that fulfills it.
+Does that imply that Inner Source projects place no value on face to face
+communication? Not quite: There is value in meeting participants face to face.
+Remember that all written communication lacks a lot of bandwidth compared with
+meeting in person: There's no gestures, no mimics, not even the tone of voice
+that otherwise help with understanding. In particular for inter-personal
+challenges, resolving conflict and misunderstandings meeting in person has
+strong value. However for Inner Source projects the goal should be that
+communication about project decisions be kept in writing so that others can
+follow along and influence the project, so that even years later it's possible
+to trace why a certain decision was made.
 
-In some organisations you may need to talk to the host team in a synchronous manner, e.g. around a coffee, by means of a video of voice chat, in addition to interacting with the project asynchronously, e.g. by means of issue trackers or email to establish some trust. 
+Here's my general rule of thumb: Feel free to meet over coffee, often that helps
+with building a stronger team. Remember though that when the team is split into
+multiple physical locations (for me that includes teams that are spread across
+mutliple floors) there is a good chance that this encourages the team to split
+into multiple silos. Do make sure though that all decisions are made in a
+transparent and asynchronous way so that everyone - including those lurking in
+on your conversation - can jump in and become active contributors. One example
+of just how far one can take open decision making is explained in several
+excercises in the [Open Organization
+Workbook](https://opensource.com/open-organization/resources/workbook)
 
-In the same way that real-life hosts don't always appreciate a stranger knocking on their door, some InnerSource hosts appreciate some heads-up before seeing a pull request gets opened to their repo. Typically there will be an issue tracker where new features are discussed before being submitted as a pull request. A useful alternative are archived, searchable, linkable but unstructured discussion platforms: Thank of mailing lists, chat plattforms or online fora.
+Now, how do you figure out where an Inner Source project would like to discuss
+changes, future direction of the project? Many InnerSource projects outline how
+they like to be approached by potential Contributors in their README.md. If that
+document becomes too large to handle contribution guidelines tend to be split
+out into a file named CONTRIBUTING.md files. Following those recommendations
+greatly helps the contributor selling their offer.
 
-In all of these interactions, be prepared to "sell" your contribution to the host team. Articulate the value that the contribution will bring to their ecosystem.
+In all of these interactions, be prepared to "sell" your contribution to the
+host team. Articulate the value that the contribution will bring to their
+ecosystem.
 
-The host team will be the one taking over maintenance for your changes. It makes sense to offer to fulfill a 30-day warranty on your submission. This can alleviate the host teams' fear of the contributors not being available for support with fixing bugs after the time on contribution.
-
-Many InnerSource projects outline how they like to be approached by potential Contributors in their README.md or CONTRIBUTING.md files. Following those recommendations greatly increases your chances of getting your contribution accepted.
+The host team will be the one taking over maintenance for your changes. It makes
+sense to offer to fulfill a [30-day warranty](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/30-day-warranty.md) on your submission. This can
+alleviate the host teams' fear of the contributors not being available for
+support with fixing bugs after the time on contribution.
 
 #### Anticipate and follow house rules
 
--- a/contributor/03-contributor-ethos.md
+++ b/contributor/03-contributor-ethos.md
@@ -26,7 +26,7 @@ In the real world there are several ways that guests are invited in:
 * Knock on the door unexpected.
 * Knock on the door only when invited.
 
-With InnerSource projects there are several ways that code gets in:
+In like manner, a contribution to an InnerSource project can have several ways that it gets accepted.
 
 Sometimes you just open a pull request. At times you may need to get a feature request accepted before opening the pull request that fulfills it.
 
--- a/contributor/03-contributor-ethos.md
+++ b/contributor/03-contributor-ethos.md
@@ -20,7 +20,7 @@ How do these expectations apply to InnerSource projects?
 
 #### Getting in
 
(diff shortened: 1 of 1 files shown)


contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+In the last segment we have motivated why you would want to reuse components and become active as a Contributor. This artice is sharing best practices on how to successfully contribute your changes to the host team's code base.
+
+ ### Contributor Ethos
+
+An InnerSource Contributor trying to make a contribution to the host team essentially is a guest in their home. In general a good guest is expected to behave in a certain way:
+
+* They knock at the door.
+* They anticipate and follow house rules.
+* They understand they are not the home owner and acty accordingly.
```
rrrutledge:
```suggestion
* They understand they are not the home owner and act accordingly.
```

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+How do these expectations apply to InnerSource projects?
+
+#### Getting in
+
+In the real world their are several ways that guests are invited in:
```
rrrutledge:
```suggestion
In the real world there are several ways that guests are invited in:
```

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+* An "open" sign lets them know they can walk right in without knocking.
+* Knock on the door unexpected.
+* Knock on the door only when invited.
+
+With InnerSource projects there are several ways that code gets in:
```
rrrutledge: Suggest tying this section to the preceding text a little more directly.

```suggestion
In like manner, a contribution to an InnerSource project can have several ways that it gets accepted.
```

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+Sometimes you just open a pull request. At times you may need to get a feature request accepted before opening the pull request that fulfills it.
```
rrrutledge: > At times you may need to get a feature request accepted before opening the pull request that fulfills it.

This point is made with more detail in the paragraph `In the same way that ...`.  Suggest just deleting this sentence here since the point will be brought up later.

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+In some organisations you may need to talk to the host team in person in addition to interacting with the project digitally.
```
rrrutledge: Suggest putting this point after the following paragraph so that the ways of getting in are ordered from lowest-touch (just open a PR) to highest touch (meet the host team first).

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+In the same way that real-life hosts don't always appreciate a stranger knocking on their door, some InnerSource hosts appreciate some heads-up before seeing a pull request gets opened to their repo. Typically there will be an issue tracker where new features are discussed before being submitted as a pull request. A useful alternative are archived, searchable, linkable but unstructured discussion platforms: Thank of mailing lists, chat plattforms or online fora.
+
+In all of these interactions, be prepared to "sell" your contribution to the host team. Articulate the value that the contribution will bring to their ecosystem.
+
+The host team will be the one taking over maintenance for your changes. It makes sense to offer to fulfill a 30-day warranty on your submission. This can alleviate the host teams' fear of the contributors not being available for support with fixing bugs after the time on contribution.
+
+Many InnerSource projects outline how they like to be approached by potential Contributors in their README.md or CONTRIBUTING.md files. Following those recommendations greatly increases your chances of getting your contribution accepted.
```
rrrutledge: I know you're just following the order in the outline, but as I read the article now I feel like this paragraph would go better. before the paragraphs on "selling".

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+#### Anticipate and follow house rules
```
rrrutledge: Is there a way to make this section a little more conversational with some flow to it?  Right now all of the information is there but it looks like just the straight text of the bullet points without any other structure to make it enjoyable to read.

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+You should understand the project well enough to know that your contribution makes sense overall and is not just your personal snowflake.
+
+
+#### Understands they are not the home owner and acts accordingly.
+
+Being a Contributor essentially means being somehow closer to the host team than someone merely requesting a feature. Still Contributors are not accountable for the software project they are contributing to.
+
+As a result the final call on what the contribution must look like is with the host team. It helps to make sure to approach the host team with a humble mindset, with the assumption that all are collaborating towards the purpose of the shared organisation. It helps to be open and transparent - not only about what was implemented how, but also why the change was needed.
+
+Usually it helps to treat feedback as a gift: Others are trying to improve your solution, saving you from trouble further down the road.
```
rrrutledge: Suggest using a definitive tone.

```suggestion
Treat any feedback as a gift: Others are trying to improve your solution, saving you from trouble further down the road.
```

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+There is a chance that the host team does not accept your contribution at all. In that case it helps to work with the team, figure out if there is a sub aspect of what you need that can be solved in their project and solve the remaining issues on your end.
```
rrrutledge: Suggest breaking into multiple sentences.

```suggestion
There is a chance that the host team does not accept your contribution at all. In that case it helps to work with the team, figure out if there is a sub aspect of what you need that can be solved in their project.
Collaborate on that sub aspect and then find another way to solve the remaining issues on your end.
```

contributor/03-contributor-ethos.md
lenucksi: In the sense of "this gives the contributing person good arguments for the sale / allows them to make a good offer"?

contributor/03-contributor-ethos.md
lenucksi: I agree, this reads like a checklist / manual. I don't instantly have any good suggestions here though.

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
+# Draft comments
+
+ Outline for this section
+
+     Contributors by definition work in teams that they are not familiar with. This section explains how to avoid common pitfalls when crossing team boundaries.
+
+ ## Article
```
lenucksi: I would suggestion to use `#` for draft comment and # for the article and then `##(###, etc.)` for the (sub-)sections as this might make it easier to read during the drafting process.
```suggestion
# Article
```

contributor/03-contributor-ethos.md
lenucksi: I like that approach. 
I do think that we should add technically mediated "in-person" conversation though, i.e. a voice/video call or the likes. Sometimes even a chat will work wonders. 
"Digital interaction" could be a clearer - mail, voice, video, issue - all of those are digital. Even an appointment to discuss f2f has digital aspects. How about using the division of synchronous vs. asynchronous conversation here? Initiating a synchronous conversation (kitchenette+coffee, phonebooth, voice, video) with someone often helped for that purpose in my observation.
Adding a review here.

```suggestion
In some organisations you may need to talk to the host team in a synchronous manner, e.g. around a coffee, by means of a video of voice chat, in addition to interacting with the project asynchronously, e.g. by means of issue trackers or email to establish some trust. 
```

contributor/03-contributor-ethos.md
lenucksi: If we remove this sentence, then the case that sometimes a simple, well written PR could solve things would go missing. Maybe just rephrasing this would help

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+* Knock on the door unexpected.
+* Knock on the door only when invited.
+
+With InnerSource projects there are several ways that code gets in:
+
+Sometimes you just open a pull request. At times you may need to get a feature request accepted before opening the pull request that fulfills it.
+
+In some organisations you may need to talk to the host team in person in addition to interacting with the project digitally. 
+
+In the same way that real-life hosts don't always appreciate a stranger knocking on their door, some InnerSource hosts appreciate some heads-up before seeing a pull request gets opened to their repo. Typically there will be an issue tracker where new features are discussed before being submitted as a pull request. A useful alternative are archived, searchable, linkable but unstructured discussion platforms: Thank of mailing lists, chat plattforms or online fora.
```
lenucksi: Just split this large paragraph up into individual lines to allow for easier reviewing. 

```suggestion
In the same way that real-life hosts don't always appreciate a stranger knocking on their door, some InnerSource hosts appreciate some heads-up before seeing a pull request gets opened to their repo.
Typically there will be an issue tracker where new features are discussed before being submitted as a pull request.
A useful alternative are archived, searchable, linkable but unstructured discussion platforms: Thank of mailing lists, chat plattforms or online fora.
```

Along the same lines according to the updated outline #99 we should definitely strongly highlight the value of keeping as much in the archived, indexed public as possible. Might even want to encourage the TCs (article) to continuously encourage such behavior as this tends to be forgotten over time in many organizations.

contributor/03-contributor-ethos.md
```diff
@@ -0,0 +1,64 @@
...
+
+In all of these interactions, be prepared to "sell" your contribution to the host team. Articulate the value that the contribution will bring to their ecosystem.
+
+The host team will be the one taking over maintenance for your changes. It makes sense to offer to fulfill a 30-day warranty on your submission. This can alleviate the host teams' fear of the contributors not being available for support with fixing bugs after the time on contribution.
+
+Many InnerSource projects outline how they like to be approached by potential Contributors in their README.md or CONTRIBUTING.md files. Following those recommendations greatly increases your chances of getting your contribution accepted.
+
+#### Anticipate and follow house rules
+
+The README.md or CONTRIBUTING.md often not only contains information about how to approach the host team, but also guidelines throughout the contribution process. Be sure to understand and follow them.
```
arnom-ms: I would like to make a greater distinction between README.md and CONTRIBUTING.md. I suggest we suggest (:-)) to use the README.md to explain what the project is about and how to USE, Build and get started with it. In contrast, the CONTRIBUTING.md  should be used to describe HOW TO CONTRIBUTE to the project (process wise, code styles, test expectations etc).  Maybe I should open an issue for review on docs in general?

contributor/03-contributor-ethos.md
MaineC: You mean something along what I talked about there maybe:

https://drive.google.com/drive/u/0/folders/0ByhNYwghjRyeOVlMN0lBYndsdXc (starting page 47 and following)?

contributor/03-contributor-ethos.md
MaineC: I think it makes sense to be a bit more specific in this paragraph, however I wouldn't strictly forbid contributing related docs in the README. More like: If you don't find contributing related documentation in the README they have been split out into CONTRIBUTING.md

The reasoning behind that: For small enough projects both sides of the documentation can be short enough to be kept in one document. Even if they are split, the contributor will still have to read the README anyway to first understand how the project works and can be used before even thinking about contributing.

Makes sense?

lenucksi: I agree with both opinions. 
To me it appears that the shared goal is to prevent too long documentation and enable easier absorption of the knowledge contained in the documents.

With larger projects separation is totally great, with the smaller ones, combining them is fine too. What we could provide is a guidance when to separate such documents and how to arrive at a (set of) document(s) that serve their purpose well. 
I was working at such documents a while ago, let's see what became of them.

contributor/03-contributor-ethos.md
rrrutledge: You've done a lot of work!  For me that link goes to a folder rather than a specific document, though.

contributor/03-contributor-ethos.md
MaineC: I've tried to re-phrase the entire paragraph. One thing I was struggling with was finding a good real world story of walking in with an open sign out up to making an appointment.  So I've changed the narrative a bit.

rrrutledge: Oh you've done some wonderful writing, but my comment was meant for the **Anticipate and follow house rules** section (rather than the **Getting in** section).

MaineC: Ups :) I'll look at the other section as well then.

PR #109, opened 2019-08-07
Fix Slack name

rrrutledge:
paypalflow -> innersourcecommons


rrrutledge:
I was trying to use the documentation and ran across a broken link.  Fixed them all, then.

PR #111, opened 2019-08-12
Rewrite introductory paragraphs

rrrutledge:
To appeal to an individual who just cares about making progress on their project.

### DIFF
--- a/contributor/01-introduction-article.md
+++ b/contributor/01-introduction-article.md
@@ -23,15 +23,12 @@ How do I best contribute to an InnerSource project?  This segment shares a summa
 
 ## The Contributor in InnerSource.
 
-Without contributions, an InnerSource community and its project(s) would not exist in a meaningful way.
-There would be open repositories, ideally equipped with somewhat specific extra documents (e.g. a CONTRIBUTING.md and a README.md) and person(s) committing to them, e.g. a Trusted Committer.
-[EDITOR NOTE:
-- In other articles we draw a distinction between guest and host team.
-- We argue that the Committer is part of the guest team and needs to fit in with the host team.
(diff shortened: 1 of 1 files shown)


contributor/01-introduction-article.md
```diff
@@ -23,15 +23,12 @@ How do I best contribute to an InnerSource project?  This segment shares a summa
...
-- We argue that the Committer is part of the guest team and needs to fit in with the host team.
-- In a more-than-one-person project what do we call the people aside of the TC/PO that are members of the host team? Just team members?
-        - OSS would likely not make this distinction and everyone would end up as contributor until it's committer / trusted committer at some point I guess?
-        - In the ASF-Rule-set this would be the distinction between Developer and Committer.
-]
+Have you ever been blocked in your next coding task because another team didn't have time to add a feature in their system that you depend on?
+Perhaps after a while you even had to do some extra work in your project to work around the missing feature.
+How nice would it be to never be blocked in this way?
+
+With projects that run via InnerSource, you'll never be blocked waiting for another team to deliver some needed feature.
```
lenucksi: From what I remember from the introduction and some places possibly in TC we've mentioned that InnerSource is a non-exclusive way to work on a project, i.e. it's not either InnerSource or Agile but some tasks might be Agile, some InnerSource - and some projects are entirely InnerSource. 
"run via InnerSource" triggers a somewhat exclusive connotation with me here - what do the others think?
And does someone who sees it the same maybe even have a different, commit suggestion-ready wording in my mind?

rrrutledge: ~run via InnerSource~ `incorporate InnerSource principles`

PR #118, opened 2019-08-21
How to write InnerSource

rrrutledge:


MaineC:
Looks good to me. +1 to spelling the term consistently and providing documentation on which version we use in the learning path.

rrrutledge:
Things sounded good on the call today so merging.

Issue #114, opened 2019-08-14
Fix header levels for all articles


MaineC:
http://docs.atlas.oreilly.com/writing_in_markdown.html#writing-markdown-for-htmlbook-Oytyf4

As noticed in the review of #101 we need to go over all articles after writing them to fix the header levels.

rrrutledge:
Thanks for opening this issue, Isabel.

lenucksi:
Results from #108 might help here.

Ludmila-N:
After reading O'Reilly description, I see that I need to align all headers to a different level - doing so.

lenucksi:
Closing this since it's in done now.

PR #181, opened 2019-11-02
addresses #114

Ludmila-N:
Aligned article headers to O'Reilly standard (H1) 
[#114 ](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/114)

### DIFF
--- a/contributor/02-becoming-a-contributor-article.md
+++ b/contributor/02-becoming-a-contributor-article.md
@@ -1,13 +1,13 @@
 
-## Becoming an InnerSource Contributor
+# Becoming an InnerSource Contributor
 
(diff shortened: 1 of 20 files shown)


contributor/02-becoming-a-contributor-article.md
```diff
@@ -1,13 +1,13 @@
 
-## Becoming an InnerSource Contributor
+# Becoming an InnerSource Contributor
 
 InnerSource contributors operate outside of regular team boundaries, they are the links crossing silos within organisations. As such they need to be aware of a few common practices that make this work more effective.
 
-### Sharing Mindset
+## Sharing Mindset
```
lenucksi: If we're parting from using title case (which the other non-header edits aim at) then this should also be corrected.
```suggestion
## Sharing mindset
```

PR #190, opened 2019-11-14
Atlas sync

lenucksi:
This GitHub Actions pipeline synchronizes our repo content on changes to master to the O'Reilly Atlas repository and merges in the local specifics to enable rendering the content properly.

As a last step a build on O'Reilly is triggered using the Gem described here: http://docs.atlas.oreilly.com/api.html

Secrets live in the project and run off of my account first. Having a bot account for that purpose probably would do no harm though.

Once this works, extending it to the other repos is a quick fix.

**TODO:** I might need to add runs using the actual `GITHUB_SHA` of the current state to enable using this as a pull request check. Or might need to add a second pipeline running against pull request checks separating the sync run on pushes to master and the pull request check run.

### DIFF
--- a/.github/workflows/SyncToAtlas.yml
+++ b/.github/workflows/SyncToAtlas.yml
@@ -42,7 +42,7 @@ jobs:
         git push oreilly oreilly-master:master
     - name: Trigger build on Atlas
       run: |
-        gem install atlas
+        gem install atlas-api
         atlas build $OREILLYAPISECRET lenucksi/learningpath-introduction pdf,epub master
       env:
         OREILLYAPISECRET: ${{ secrets.OREILLYAPISECRET }}
--- a/.github/workflows/SyncToAtlas.yml
+++ b/.github/workflows/SyncToAtlas.yml
@@ -1,0 +1,48 @@
+name: Sync Master to Atlas Repos
+
+on: 
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  atlas:
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@master 
+    - name: Add O'Reilly remotes
+      run: |
+        git remote add oreilly git@git.atlas.oreilly.com:lenucksi/learningpath-introduction.git
+    - uses: fregante/setup-git-token@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+    - name: enable SSH-Key
+      env: 
+        OREILLYKEY: ${{ secrets.OREILLYKEY }}
+      run: |
+        mkdir -p $HOME/.ssh
+        #ls -al ~/.ssh
+        echo -e "$OREILLYKEY" > ~/.ssh/id_rsa
+        chmod 600 ~/.ssh/id_rsa
+        eval `ssh-agent -s`
+        ssh-add ~/.ssh/id_rsa
+        ssh-keyscan git.atlas.oreilly.com >> ~/.ssh/known_hosts
+        #    - name: checkout master
+        #      run: git checkout master
+    - name: push master to oreilly repo
+      run: |
+        # What will happen here?
+        # We've got the repo local and need to merge two histories.
+        # O'Reilly master has extra content, not conflicting with ours though. Fetch it and check it out
+        # Get our master into a different branch. Merge our master-branch into the O'Reilly master.
+        # Push our new merged master into the O'Reilly master.
+        # Party!
+        git fetch oreilly
+        git checkout -b oreilly-master oreilly/master
+        git merge -m "Merge GitHub content to Atlas" -v --commit --log master 
+        git push oreilly oreilly-master:master
--- a/.github/workflows/SyncToAtlas.yml
+++ b/.github/workflows/SyncToAtlas.yml
@@ -46,3 +46,9 @@ jobs:
         git checkout -b oreilly-master oreilly/master
         git merge -m "Merge GitHub content to Atlas" -v --commit --log master 
         git push oreilly oreilly-master:master
+    - name: Trigger build on Atlas
+      run: |
+        gem install atlas
+        atlas build $OREILLYAPISECRET lenucksi/learningpath-introduction pdf,epub master
+      env:
+        OREILLYAPISECRET: ${{ secrets.OREILLYAPISECRET }}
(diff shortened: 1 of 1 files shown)


.github/workflows/SyncToAtlas.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@master 
+    - name: Add O'Reilly remotes
+      run: |
+        git remote add oreilly git@git.atlas.oreilly.com:lenucksi/learningpath-introduction.git
+    - uses: fregante/setup-git-token@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
```
rrrutledge: Where are these secrets stored?

.github/workflows/SyncToAtlas.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: enable SSH-Key
+      env: 
+        OREILLYKEY: ${{ secrets.OREILLYKEY }}
+      run: |
+        mkdir -p $HOME/.ssh
+        #ls -al ~/.ssh
+        echo -e "$OREILLYKEY" > ~/.ssh/id_rsa
+        chmod 600 ~/.ssh/id_rsa
+        eval `ssh-agent -s`
```
rrrutledge: Just curious - why is `eval` needed here instead of just running `ssh-agent -s` directly?

.github/workflows/SyncToAtlas.yml
```diff
@@ -0,0 +1,54 @@
...
+        ssh-add ~/.ssh/id_rsa
+        ssh-keyscan git.atlas.oreilly.com >> ~/.ssh/known_hosts
+        #    - name: checkout master
+        #      run: git checkout master
```
rrrutledge: These commented-out lines no longer needed?

.github/workflows/SyncToAtlas.yml
```diff
@@ -0,0 +1,54 @@
...
+jobs:
+  atlas:
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@master 
+    - name: Add O'Reilly remotes
+      run: |
+        git remote add oreilly git@git.atlas.oreilly.com:lenucksi/learningpath-introduction.git
```
rrrutledge: Is this some fork of the actual learning path?

.github/workflows/SyncToAtlas.yml
```diff
@@ -0,0 +1,54 @@
...
+        # Get our master into a different branch. Merge our master-branch into the O'Reilly master.
+        # Push our new merged master into the O'Reilly master.
+        # Party!
+        git fetch oreilly
+        git checkout -b oreilly-master oreilly/master
+        git merge -m "Merge GitHub content to Atlas" -v --commit --log master 
+        git push oreilly oreilly-master:master
+    - name: Trigger build on Atlas
+      run: |
+        gem install atlas
```
rrrutledge: This is [Hashicorp Atlas](https://rubygems.org/gems/atlas/)?  Is that related to O'Reilly Atlas?

StingRayZA: I think this should be `gem install atlas-api`...
http://docs.atlas.oreilly.com/api.html#atlas-gem

.github/workflows/SyncToAtlas.yml
StingRayZA: @rrrutledge they're stored in the project itself. this particular one is a builtin - but the others can be found by navigating to Settings | Secrets on the project...

.github/workflows/SyncToAtlas.yml
StingRayZA: `ssh-agent` outputs some env variables when it runs. if you wrap it in an `eval`, it activates those and makes the agent available to subsequent calls.
https://linux.die.net/man/1/ssh-agent (look under _There are two main ways to get an agent set up_)

.github/workflows/SyncToAtlas.yml
lenucksi: Exactly what @StingRayZA says.

.github/workflows/SyncToAtlas.yml
lenucksi: Yes, indeed.

.github/workflows/SyncToAtlas.yml
rrrutledge: Thanks.

.github/workflows/SyncToAtlas.yml
rrrutledge: Got it.

.github/workflows/SyncToAtlas.yml
rrrutledge: Thanks.

.github/workflows/SyncToAtlas.yml
rrrutledge: Thanks.  Will anyone else ever need to see what is in _Atlas_?  It looks like I, myself, can't view this repo, for example.

.github/workflows/SyncToAtlas.yml
rrrutledge: Yes!  https://rubygems.org/gems/atlas-api/

.github/workflows/SyncToAtlas.yml
lenucksi: > Thanks. Will anyone else ever need to see what is in _Atlas_? 

Aside from people at O'Reilly? No. Some of it is likely O'Reilly IP anyway.

> It looks like I, myself, can't view this repo, for example.

Now you should be able to.

.github/workflows/SyncToAtlas.yml
lenucksi: Thanks for catching that! Looks like a case of incomplete commandline copypasta.
```suggestion
        gem install atlas-api
```

.github/workflows/SyncToAtlas.yml
lenucksi: > Thanks. Will anyone else ever need to see what is in _Atlas_? 

Aside from people at O'Reilly? No. Some of it is likely O'Reilly IP anyway.

> It looks like I, myself, can't view this repo, for example.

Now you should be able to.

Issue #157, opened 2019-10-17
Trusted Committer: synchronize article markdowns from O'Reilly back into github


Ludmila-N:
We've been informed that O'Reilly editors have made their changes to the Trusted Committer article, and [we have the results in pdf format](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/58).
We are still [making content changes](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/41) to the article and need to sync the final markdowns from O'Reilly back to this project.

rrrutledge:
Thanks for opening, Ludmila!

lenucksi:
#166 tracks needed format conversion

lenucksi:
Since #195 is merged now all our content is in ASCIIdoc format. 
Also, our repo now is the upstream for O'Reilly too and the Atlas repositories are downstream repos since the successful closing of #108.

Thus we can manually merge the already existing edits in the Atlas ASCIIdoc repositories into our repository and solve that way.
We have edited content for the Trusted Committer Segment and the Introduction segment already.

rrrutledge:
Some edits are merged on, others still pending.

rrrutledge:
https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/262 is the last one.

Ludmila-N:
Closing this issue as all relevant PRs have been merged.

rrrutledge:
Hooray!  Excellent!

Issue #58, opened 2019-03-18
Publish Written Articles


rrrutledge:
Now that the [Trusted Committer Videos](https://learning.oreilly.com/videos/the-trusted-committer/9781492047599) are out there we need to publish the [written articles](https://github.com/InnerSourceCommons/InnerSourceLearningPath/tree/master/trusted-committer) as well.  We'd like each article to occupy its own page in the Learning Path.

rrrutledge:
When the PDFs of the articles are available you can just attach them to this issue and move to `In Review`.

rrrutledge:
@clobner I followed up on who should be listed as author on the _Trusted Committer_ written articles.

I looked through the _Git_ history and it @gruetter and I are the only committers to the articles.  We talked about it and Georg should be listed as the primary author and myself secondary.

lenucksi:
From the point of initially reviewing the material living in all the branches, forks and merge request discussions I think there might have been additional reviews / contributions from others to the TC articles that we might want to give credit to in a way. 
Especially review-contributions from times before our use / the existance of the commit suggestion feature might be relevant here - hoping that the CS feature would ensure attribution for current work.
For example, I've found 42fb5521f2ab62ed72a728406d887a6faf4ab626 crediting Cedric and Loren for contributions to the TC article.

I know that this might be a quite a bit of work, but wanted to bring it on. After all praising participants and giving credit is one of our patterns...

rrrutledge:
Sure!  Suggestions?

clobner:
Articles are with the O'Reilly content team for review.

lenucksi:
Went through the already merged PRs on the Trusted Committer and found these review contributions which appear to have had rather large effects on the resulting document.
- Cedric Williams, Noah Cawley #1 
- Guy W. Martin, Russ Rutledge #26 
- Loren Sanz, Cedric Williams, Erin MB #18 
- Erin MB, Tim Yao, Loren Sanz, Cedric Williams #9 (Reviews, pretty active PR)

> Sure! Suggestions? I know that you yourself have been a prolific reviewer in general though I can't recall how far along was the Trusted Committer work when you came on board?

The first PR activity I found was in January / February with #1 and #35 has been created due to me finding a typo on the slides (I think that's been the result of my initial review process).

rrrutledge:
Great!  Per our previous conversations we should reach out to each to see if they'd like to be listed as a reviewer.

[...]

PR #204, opened 2020-01-06
Raw, edited content from the Atlas TC repo.

lenucksi:
## Content
This pull request represents the editorial content in the Trusted Committer repository in Atlas. 
It leaves out the material such as the TOC, the front matter and build configuration.

There are no content edits made to it by me. 
There are many great edits by the O'Reilly editors, however also some parts where our codebase possibly has things that might be worth to incorporate or others that we might want to change.

This handles #157.
Added @gruetter for review since the text has been written by him to a relevant degree. 

## Todo 
If we want to use this to enable automatic sync of Upstream to Atlas and final review on GitHub we'll need to incorporate / merge these changes with our repo dealing with above differences.

- Add decent sync points to enable decent web diffing. 
  - See [here](https://github.com/InnerSourceCommons/InnerSourceLearningPath/compare/atlas/tc-edited-sources?expand=1#diff-1ff22a5a496ebe237f53da680d9c616bL45) and [here](https://github.com/InnerSourceCommons/InnerSourceLearningPath/compare/atlas/tc-edited-sources?expand=1#diff-1ff22a5a496ebe237f53da680d9c616bR50) for an example. Dealt with by adding matching commits with the proper line breaks etc..
- Review the cross chapter linking and figure out the correct way to do this for O'Reilly.
 - If we link external content, such as the patterns, figure out a way to link them properly 
- Review the actual content
- Check out the build results (PDF/Epub) to enable the final sign-off for O'Reilly to put them on the platform. See #58 for this.

### DIFF
--- a/trusted-committer/01-introduction.asciidoc
+++ b/trusted-committer/01-introduction.asciidoc
@@ -20,6 +20,7 @@ explain why we think the name is both apt and important.  Let's
 start with the https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributor_] role. A _Contributor_ - as the name implies -
 makes contributions to an InnerSource community.  These contributions could be code or non-code
 artifacts, such as bug reports, feature requests or documentation.
+
 _Contributors_ might or might not be part of the community. They might be sent by
 another team to develop a feature that their team needs. This is why we
 sometimes also refer to _Contributors_ as _Guests_ or being part of a _Guest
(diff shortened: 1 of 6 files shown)


rrrutledge:
Thanks, Johannes!  Will look at this next week.

rrrutledge:
It's too bad there isn't a cleaner diff so that we can see exactly what were the changes from O'Reilly.  It's nice to have this content, though.

lenucksi:
> It's too bad there isn't a cleaner diff so that we can see exactly what were the changes from O'Reilly. It's nice to have this content, though.

That's exactly what adding a few changes to create better sync points for the GitHub diff algorithm is supposed to help us with.

lenucksi:
Closing this for now in favor of #205.

PR #205, opened 2020-01-07
Atlas/tc edited sources 2

lenucksi:
Similar to #204 but using different sync points and mergeable.

Unfortunately the GitHub web UI does not do diffing on this as good as the usual tools like Meld. 
Updating this once I've got something better.

Handles #157


rrrutledge:
Thanks for this work!  Will try to look at this in the upcoming week.

lenucksi:
Closing in favor of #211 etc. merging in separate PR's

PR #208, opened 2020-01-17
Remove tip about creating a conference call

rrrutledge:
We just have a single conference call for the whole series now.


rrrutledge:
Thank you!

PR #210, opened 2020-01-28
Atlas changes to TC 01-intro article

Ludmila-N:
This PR is for the introduction article only.
Changed TC back to Trusted Committer; re-added the links. Kept their style suggestions.


Ludmila-N:
will redo and recreate

PR #212, opened 2020-02-16
Sync Atlas changes made in TC "02-ensuring-product-quality" article

Ludmila-N:
Accepted most style changes from Atlas; reintroduced links and essential terms back to what we wanted them to be.
Addresses #157


Ludmila-N:
@clobner - please help with one specific change.
I'm not 100% clear on this sentence in the second paragraph in "02-ensuring-product-quality.asciidoc", towards the end (this is from Atlas):
`A high level of software quality is essential for establishing and maintaining the trust of both users in the software the community is building and the management in the way it is being built.` 

I partially reverted it to the original way:
`A high level of software quality is essential for establishing and maintaining the trust in the InnerSource community on part of their users and their management.`

What we meant is that InnerSource community wants to be trusted by (1) users of their software, and (2) managers of the Trusted Committers. If my version is not good enough, I'll be happy to accept a fix, so long as it delivers the same meaning, and does so in a reasonably simple way.
Appreciate your help!

PR #215, opened 2020-02-23
Sync 03-keeping-the-community-healthy from Atlas

Ludmila-N:
Addresses #157

### DIFF
--- a/trusted-committer/03-keeping-the-community-healthy.asciidoc
+++ b/trusted-committer/03-keeping-the-community-healthy.asciidoc
@@ -45,7 +45,7 @@ We believe that Trusted Committers should prioritize onboarding and mentoring du
 requests over reaching communicated release dates, unless there is a very
 good reason not to. Good mentoring during pull requests leads to a higher level
 of trust and engagement by https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_], which in turn leads
-to more contributions. We’ll discuss this more in pass:[<a data-type="xref" href="upleveling" data-xrefstyle="chap-num-title">#upleveling</a>].
+to more contributions. We’ll discuss this more in https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/trusted-committer/04-uplevelling-community-members.asciidoc["Upleveling the Community"].
 
 Finally, some people stick around in InnerSource communities because
 they get to focus on developing software instead of activities considered overhead or waste, especially 
@@ -62,15 +62,11 @@ in identifying and solving problems. Ultimately, Trusted Committers should empow
 community to find process-related problems and to adapt and improve
 them as a community as much as possible.
 
-For Trusted Committers to be able to fulfill all these responsibilities, it is important that
-they communicate regularly with community members and _keep an ear to the
-ground_ so to speak, so that they are aware of the community's needs. We'll
-go into more detail about this in the section in https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/trusted-committer/06-advocating-for-the-communitys-needs.asciidoc["Advocating the Community's
-Needs"].
-
 For Trusted Committers to be able to fulfill all these responsibilities, it is
 important that they communicate regularly with community members and
-keep an ear to the ground. We’ll go into more detail about this in pass:[<a data-type="xref" href="advocating" data-xrefstyle="chap-num-title">#advocating</a>].
+keep an ear to the ground.  We'll
+go into more detail about this in the section in https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/trusted-committer/06-advocating-for-the-communitys-needs.asciidoc["Advocating the Community's
+Needs"].
 
 In summary, Trusted Committers should strive to create a welcoming and appreciative
 environment for their https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_] that allows them to concentrate on writing
--- a/trusted-committer/03-keeping-the-community-healthy.asciidoc
+++ b/trusted-committer/03-keeping-the-community-healthy.asciidoc
@@ -34,7 +34,9 @@ and to grow personally. This is again where the role of the Trusted Committer is
 important. Trusted Committers often become mentors for junior developers, and
 explicitly spend time during pull requests not only pointing out areas
 for improvement but also explaining in detail why something needs to be
-improved and how to do it. By doing so, Trusted Committers
+improved and how to do it. 
+They supply the theory or experience behind the change and offer suggestions for the best ways to do it
+By doing so, Trusted Committers
 can increase the speed of learning in their
 communities far beyond that in traditional software
 development projects.
--- a/trusted-committer/03-keeping-the-community-healthy.asciidoc
+++ b/trusted-committer/03-keeping-the-community-healthy.asciidoc
@@ -1,72 +1,64 @@
-== Keeping the community healthy
+== Keeping the Community Healthy
 
-The introduction points out that Trusted Committers have both tech oriented and community
-oriented responsibilities. It is not sufficient to focus on code and code
-health only. To ensure success in the long run, Trusted Committers should strive for keeping
-the community which is building the software healthy as well. Because of that, they
-must strike a good balance between effort spent on ensuring product quality and
-growing a healthy community.
+The introduction pointed out that Trusted Committers have both tech-oriented and
+community-oriented responsibilities. It is not sufficient to focus on
+code and code health only. To ensure success in the long run, Trusted Committers should
+strive to keep the community building the software healthy
+as well. Because of that, they must strike a good balance between ensuring product quality and growing a healthy community.
 
 What does a healthy community look like? Quite simply, in a healthy community,
-https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_] tend to stick around, can spend most of their time on developing
-software and are able to level up their abilities. As a result, a healthy
-community will also be growing.
+https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_] tend to stick around, can spend most of their time developing software, and are able to level up their abilities.
+As a result, a healthy community will be continually growing.
 
 Why do https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_] join and stick around in a community? Some do because they
 subscribe to the purpose or the mission of the community. It is the Trusted Committer's job to
 clearly articulate and promote this purpose. The importance of this is often
-not recognized - marketing a community and their products is truly essential.
+not recognized, but marketing a community and its products is truly essential.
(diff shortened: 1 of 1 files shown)


Ludmila-N:
Attached the mapping of the links that differ between github and Atlas in the first 3 articles
[ISC-learning-path-links.txt](https://github.com/InnerSourceCommons/InnerSourceLearningPath/files/4240459/ISC-learning-path-links.txt)

trusted-committer/03-keeping-the-community-healthy.asciidoc
```diff
@@ -1,80 +1,76 @@
...
-Trusted Committers are the ones who should create and enact a code of conduct for the
-community. Should there be community members whose behavior is detrimental or
-even toxic to the community health, it is the Trusted Committer's responsibility to either
-try and change or contain this or, in the worst case, to remove people from the
-community. Trusted Committers should create opportunities for people to get together
-regularly and get to know each other personally.
+Another, more obvious reason for people to stick around is that they
+enjoy working with other members of the community, including the Trusted Committers. A thriving community is one where members treat
+and communicate with each other with the utmost respect. Contributions are
+treated like gifts or donations, and excellent (especially
```
lenucksi: I liked the mention of "rather than something that detracts from their own work" - this is something I have repeatedly seen leading to bulk rate handling of the contribution and hence more often than not failure of the relationship / attempt to contribute, especially on first contributions where there's no pre-established rapport.
 
I'm open for a different phrasing, but giving this difference is valuable in my opinion.
Academic publications call this "prioritized mentorship" if I remember correctly and list it as one of the pillars of working InnerSource..

trusted-committer/03-keeping-the-community-healthy.asciidoc
```diff
@@ -1,80 +1,76 @@
...
+first) contributions are lauded. The Trusted Committer’s job in all this is primarily to set an
+example for others, similar to setting an example for the level of
+software quality expected. If necessary, the Trusted Committers are the ones
+who should create and enact a code of conduct for the community. If
+there are community members whose behavior is detrimental or toxic to
+the community’s health, it is the Trusted Committer’s responsibility to address this.
```
lenucksi: The original text has specific advice on options here (attempt to fix, contain, remove) which the changed text does not have anymore. 
Do we have any rationale for keeping or removing this?

trusted-committer/03-keeping-the-community-healthy.asciidoc
```diff
@@ -1,80 +1,76 @@
...
+and communicate with each other with the utmost respect. Contributions are
+treated like gifts or donations, and excellent (especially
+first) contributions are lauded. The Trusted Committer’s job in all this is primarily to set an
+example for others, similar to setting an example for the level of
+software quality expected. If necessary, the Trusted Committers are the ones
+who should create and enact a code of conduct for the community. If
+there are community members whose behavior is detrimental or toxic to
+the community’s health, it is the Trusted Committer’s responsibility to address this. 
+Trusted Committers should create opportunities for people to get together
+regularly and get to know each other personally and to peacefully resolve conflicts as they arise.
```
lenucksi: I appreciate the "resolve conflicts" part here, however one could read it as if resolution of conflicts is only possible in person or should be done in person. 
It can in fact often help to do so, however is in my view not mandatory. Or even possible in many remote / distrib conditions.

trusted-committer/03-keeping-the-community-healthy.asciidoc
```diff
@@ -1,80 +1,76 @@
...
-so, the Trusted Committers will often be able to increase the speed of learning in their
-communities far beyond what is common in traditional software development
-projects.
+People also tend to stick around because working in an
+InnerSource community is an excellent opportunity to acquire new skills
+and to grow personally. This is again where the role of the Trusted Committer is really
+important. Trusted Committers often become mentors for junior developers, and
+explicitly spend time during pull requests not only pointing out areas
+for improvement but also explaining in detail why something needs to be
+improved and how to do it. By doing so, Trusted Committers
```
lenucksi: I liked the original, split up sentence better here since it appears more mentoring focused to me.

trusted-committer/03-keeping-the-community-healthy.asciidoc
```diff
@@ -1,80 +1,76 @@
...
-One important aspect of these is to explain what I
-call _signaling_ in pull requests: what should a comment look like? What does
-it mean if I _like_ or _+1_ a comment?  How is @mentioning someone with a /CC
-prefix different from doing so with a /FYI prefix? Generally speaking, Trusted Committers need
-to make sure that the contribution process uncovers problems, rather than being
-the cause of them.  Ultimately, Trusted Committers should empower their community to point out
-process related problems and to adapt and improve them as a community as much
-as possible.
+One important aspect of these guidelines is to explain what we call _signaling_ in
+pull requests: what should a comment look like? What does it mean if I
```
lenucksi: Should the first what not be capitalized too?

lenucksi:
> Attached the mapping of the links that differ between github and Atlas in the first 3 articles
> [ISC-learning-path-links.txt](https://github.com/InnerSourceCommons/InnerSourceLearningPath/files/4240459/ISC-learning-path-links.txt)

Thanks for starting this. Could you please add this as a file on the repo, either as a separate pull request or even within this one? 
That way anyone could easily extend it or turn it into some format than can be parsed which is what we'll need in the end.

trusted-committer/03-keeping-the-community-healthy.asciidoc
Ludmila-N: @lenucksi , I do not see any issues with capitalization in this sentence. If you mean "what" in "what should a comment look like?", the answer is no - as written, the "what" after the colon should begin with lower case. Please let me know if you meant something else.
Thank you!

trusted-committer/03-keeping-the-community-healthy.asciidoc
Ludmila-N: > Contributions are treated as gifts or donations, rather than something that detracts from `their` own work...

It's not very clear in this case who "they" are, as TCs are last mentioned two sentences prior.
@lenucksi - will this version work, in your opinion?

> Contributions are treated like gifts or donations `rather than distractions`, and excellent (especially first) contributions are lauded.  

Please let me know, I will adjust accordingly.

trusted-committer/03-keeping-the-community-healthy.asciidoc
rrrutledge: Could have a period instead of a colon and then it will be obvious to capitalize the `W`.

rrrutledge:
> Could you please add this as a file on the repo, either as a separate pull request

See [URL Mapping Table](https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/220).

trusted-committer/03-keeping-the-community-healthy.asciidoc
lenucksi: Can you add your suggestions as [commit suggestions](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request)? That way it is easier to see how they will work. They are available via the +/- symbol in the toolbar here, i.e. the first icon.

Ludmila-N: Suggested change to line 21: 
Contributions are treated like gifts or donations rather than distractions, and excellent (especially first) contributions are lauded.

trusted-committer/03-keeping-the-community-healthy.asciidoc
Ludmila-N: Assuming we are talking about the line "it is the Trusted Committer's responsibility to either try and change or contain this or, in the worst case, to remove people from the community."
I actually like O'Reilly version better. 
1 - "try and change or contain this" is neither grammatically clear nor stylistically eloquent
2 - IMHO "to address" gives people freedom to act on the unwanted behavior in the way they find most useful. This article is not a manual in people leadership, and I suggest we do not give prescriptive advise in this area.
3 - probably somewhat repeating what I said in p.1 - I don't know what "try AND change" means...

trusted-committer/03-keeping-the-community-healthy.asciidoc
Ludmila-N: I see this as a continuation of the "address conflicts" line with an attempt to show an example of addressing conflicts.
To your immediate question, "get together regularly and get to know each other personally" does not mean actually meeting "in person", aka being physically in the same location. Our weekly Learning Path calls are a great example of what I believe is expressed in this line.
I agree that meetings are not strictly necessary when working on the code, but I will be the first one to call a meeting as soon as I hear anyone complaining on another person's behavior. And I will start that meeting with a great deal of ice breaker, have people talk about themselves, then about their daily work, and only THEN about the conflict.

TL;DR: 
(1) I like the line as it is. 
(2) Specifically, "getting together" does not mean "in person". 
(3) If I convinced you, please confirm, otherwise please let me know what you'd like to change.

trusted-committer/03-keeping-the-community-healthy.asciidoc
Ludmila-N: I see your point. Suggested version for line 37:
Trusted Committers often become mentors for junior developers, and explicitly spend time during pull requests not only pointing out areas for improvement but also explaining in detail why something needs to be improved. `They supply the theory or experience behind the change and offer suggestions for the best ways to do it`.

trusted-committer/03-keeping-the-community-healthy.asciidoc
lenucksi: Hm, good points indeed. 
1) yup
2) I'm was hoping for people to have good judgement and not intentionally be dicks...yes, I know. ;)  
3) Good points. Do we leave it as is, e.g. add the options I mentioned in parentheses to the O'Reilly version or just take the O'Reilly version?

trusted-committer/03-keeping-the-community-healthy.asciidoc
lenucksi: I've changed this to a commit suggestion making it easier to see the actual difference. 
I think this phrasing is good. I thought about how to explicitly include the priority treatment/inspection of contributions - but figured that the metaphor of the gift hopefully should work with everyone.

trusted-committer/03-keeping-the-community-healthy.asciidoc
lenucksi: I am entirely on board with you here. Especially also with the conflict resolution approach. 
Since we aim this at people who might not be excessively used to all this remote / "not necessarily f2f" stuff I think we should add a little hint that the "personally" part can also be remote.
I have added a commit suggestion here. 

```suggestion
regularly and get to know each other personally and to peacefully resolve conflicts as they arise.
This can, but does not have to be, in the same physical location. Even regular video or even audio-only calls can have the desired positive effects.
```

trusted-committer/03-keeping-the-community-healthy.asciidoc
lenucksi: I like your change and turned it into a commit suggestion. 

Never mind the _over 9000_ line breaks, it will render fine.
We'll need to fix the line breaks later anyway throughout the entire material...

trusted-committer/03-keeping-the-community-healthy.asciidoc
Ludmila-N: I'm leaving O'Reilly version.

PR #220, opened 2020-02-27
URL mapping table

rrrutledge:
* Proposed file format for mapping table.
* Sparsely-populated now, but we can add in other entries as we:
  - create dedicated, per-segment landing pages on innersourcecommons.org.
  - publish articles in O'Reilly.
* If we like this format then can add in entries for all of the other segments of the learning path.
* As a note: I thought about grouping entries by section (Introduction/Contributor/etc.) but couldn't really see value in doing it so left that out.


rrrutledge:
> use a dictionary instead of the main list and have the keys be the sections

I thought about doing this but conscientiously opted out of doing it.  I don't think that URL replacement should have an opinion about content hierarchy and so wanted to leave that coupling out of the implementation of this config file.

As a thought experiment, if we wanted to swap around the logical hierarchy of our content would it make sense for the URL mapping table to require an update?  I'd say "no", since the mapping is all direct, 1-to-1 and doesn't care about the hierarchal position of the content.

I did intentionally couple videos and articles together because we've already seen things like places where a particular article isn't available we fall back to the corresponding video link, so that association is useful to have in the structure of this file.

arnom-ms:
I like the setup and I like the simple approach here as I should deal well with changes down the road, Changes to locations (wholesale move to different directories can be done by simple txt replace. What I suggest to add is a top/header section that explains what the file is for as it is not that obvious why it is needed/what it does.

rrrutledge:
> there is a third type of links in O'Reilly content

Good catch!  I hadn't seen that.  Do we know what does it mean?  It looks like another style of link to an article.

rrrutledge:
> a top/header section

The cleanest way to do this would be a comment, but as JSON doesn't support comments then we would need to switch to YAML.

Or we can add in the comment as part of the JSON data.  Something like:

```
{
  "description": " This data is for ...",
  "data": [
    ...
  ]
}
```

lenucksi:
> > a top/header section
> 
> The cleanest way to do this would be a comment, but as JSON doesn't support comments then we would need to switch to YAML.
> 
> Or we can add in the comment as part of the JSON data. Something like:
> 
> ```
> {
>   "description": " This data is for ...",
>   "data": [
>     ...
>   ]
> }
> ```
Inline dead key in JSON works for me. Having a separate text file next to it would be fine for me too.
Contents can probably be started with [this comments](https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/220#pullrequestreview-366412431) contents.

lenucksi:
> > use a dictionary instead of the main list and have the keys be the sections
> 
> I thought about doing this but conscientiously opted out of doing it. I don't think that URL replacement should have an opinion about content hierarchy and so wanted to leave that coupling out of the implementation of this config file.

One could add a field to a list element associating it with its current larger product. That way @rrrutledge valid points would be kept while the info is still present.

Being the functional programming person that I am if I touch code there's nothing in a nested dictionary that a bunch of combinators like `flatMap` and a copious amount of `collect` couldn't fix... XD

config/urls.yaml
```diff
@@ -0,0 +1,34 @@
...
+  article:
+    # URL when used as an anchor link in a published PDF.
+    anchor: ''
+    # URL in GitHub.
+    github: https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/introduction/01-introduction.asciidoc
+    # URL on the innersourcecommons.org site.
+    isc: ''
+    # URL on the O'Reilly learning plarform.
+    oreilly: ''
+- section: introduction
```
lenucksi: We could add the ordering of the elements here as e.g.
```suggestion
- section: introduction-2
```
(and the same with `introduction-1`)
Good to merge without it, too.

config/urls.yaml
```diff
@@ -0,0 +1,34 @@
+
```
lenucksi: YAMLLINT does not like blank lines at the start.
```suggestion
```

config/urls.yaml
```diff
@@ -0,0 +1,34 @@
...
+# This file contains an array of objects,
+# each having sets of URLs to the equivalent article/video in separate hosting locations.
+# This information is useful when mapping or converting one URL to another,
+# such as when generating a version of an article to be hosted on a specific platform
+
```
lenucksi: YAMLLINT suggests --- as "document start" here.
```suggestion
---
```

config/urls.yaml
rrrutledge: I'm thinking of merging without it.  They are all part of the `introduction` section.  The order is already captured by the fact that these are in an array.

PR #225, opened 2020-03-09
Sample Chinese translated article

rrrutledge:
* Trying it out to see how we like it.
* Text courtesy of a user in the InnerSource Commons.
* I do not speak, read, or write Chinese.  If any text here is inappropriate I am wholly unawares and apologize.
* Should "InnerSource" stay in English?


tanzhongyibidu:
I suggest that keep using "InnerSource" as its branding and term,
not "内部开源治理“。

rrrutledge:
That makes sense, @tanzhongyibidu.  As I am not a native speaker, you make some commit suggestions that I can merge?

lenucksi:
Thanks for starting this to the anonymous author and thanks for taking a look @tanzhongyibidu. Unfortunately I also can not speak, understand, write or read a single bit of Chinese. @rrrutledge's suggestions sound like a good plan to me.

tanzhongyibidu:
OK， I will do it

rrrutledge:
I'd love for a native speaker to sign off (@tanzhongyibidu?)

rrrutledge:
> Please tell us when you think the content is ready to merge

Ready!

Issue #226, opened 2020-03-13
Align new-line regimen through our articles


lenucksi:
Our CONTRIBUTING.md contains the agreement to have one sentence per line to enable easier reviews and restructuring. 
Much of our article material, e.g. Trusted Committer or Contributor does not adhere to that scheme.

## To Do:
- Bring up pull requests that align the material with our CONTRIBUTING.md guideline prior to O'Reilly's review process.

PR #230, opened 2020-03-22
Sync 05-lowering-the-barriers-to-entry from Atlas

Ludmila-N:
This article had been changed _**A LOT**_ by O'Reilly. 
I have put back whole paragraphs where the editor lost the original idea.
There are still significant changes left. 
In some places I want more than one pair of eyes to consider the merits of these alterations.
Feel free to request reversals if you believe the original text delivers more meaning which was lost in editing.

### DIFF
--- a/trusted-committer/05-lowering-the-barriers-to-entry.asciidoc
+++ b/trusted-committer/05-lowering-the-barriers-to-entry.asciidoc
@@ -1,75 +1,62 @@
-== Lowering the barriers to entry
+== Lowering the Barriers to Entry
 
-Soliciting contributions is one of the things that is more challenging in
-InnerSource compared to Open Source. There are a number of reasons for this.
+Soliciting contributions in an InnerSource community is more challenging than it is in an Open Source community for a number of reasons:
 
-* The sheer number of potential https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_] is lower in InnerSource
-* Contributors will want to contribute during their work time. That means
-they are more time constrained compared to doing Open Source, which also
-happens after office hours.
-* Work in InnerSource might not necessarily be part of the official
-performance goals of Contributors so time spent working on InnerSource might
-detract from reaching these goals.
+* The sheer number of potential https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_] is lower in InnerSource communities.
+* Contributors will want to contribute during their work time, which means they are more time constrained.
+* Work in InnerSource might not necessarily be part of Contributors' official
+performance goals, so time spent working on InnerSource
+may seem to detract from achieving those goals.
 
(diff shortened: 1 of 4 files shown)


lenucksi:
> This article had been changed _**A LOT**_ by O'Reilly.
> I have put back whole paragraphs where the editor lost the original idea.
> There are still significant changes left.
> In some places I want more than one pair of eyes to consider the merits of these alterations.
> Feel free to request reversals if you believe the original text delivers more meaning which was lost in editing.

I just took a look - this PR looks quite a bit more "involved" than the others. 
Thanks for making it though, also re-inserting content that got lost.

Could you do (some) of the following things to make editing this easier?
- Mark which parts you re-inserted into the PR? If you ended up committing the O'Reilly original first and then adding changes probably hinting at the right commit hashes (e.g. 7fa2e33) would do the job here.
- If possible, adjust the line breaks in the new material such that the edit view is a bit easier to go through.

Ludmila-N:
How exactly do you want the line breaks? I can do whatever is required, tell me what is needed please.

rrrutledge:
> How exactly do you want the line breaks?

My favorite is one sentence per line, one blank line between paragraphs.

lenucksi:
> How exactly do you want the line breaks? I can do whatever is required, tell me what is needed please.

In whatever way will lead to a proper GitHub review-view alignment. You'll likely have to play around a bit with newlines / blank lines.
This is just to facilitate the review for a this article. I think having the re-inserted original parts marked is more important though...

Other than that: The final result should be as @rrrutledge said: One line per sentence, one blank line between paragraphs. We can only do this once we've merged everything or else doing so will make reviewing harder. See also #226 and our CONTRIBUTING.md file.

trusted-committer/05-lowering-the-barriers-to-entry.asciidoc
```diff
@@ -1,75 +1,62 @@
...
-environment to keep the barriers to contributing as low as possible to get as
-many https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_] as possible. Trusted Committers therefore make sure that users and
-https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/contributor/01-introduction-article.asciidoc[_Contributors_] have both access to helpful documentation and people in the
-community to answer any questions they might have and that they can get up
-and running in no time. Generally speaking, Trusted Committers should make sure that the
-the onboarding experience is a great one.
+Your Contributors will often have questions about the contribution process or about the community itself, and someone has to be available to answer these questions. 
+It is therefore important for any InnerSource community to
+have one or more contact persons that are available for answering such
+questions. The Trusted Committer is usually that contact person, or else they need to make sure there is a community member "on call".
```
lenucksi: This sounds to me like there would only be one Trusted Committer, however there can be multiple. 
Putting up a suggestion here.
```suggestion
questions. Someone from the group of Trusted Committers is usually that contact person, or else they need to make sure there is a community member "on call".
```

Issue #214, opened 2020-02-21
YouTube video post-processing


lenucksi:
We've now got all videos in our hands and I've uploaded them to our YT account. They still need some post-processing for us to set them public.
Here's a copy of [the message containing the remaining tasks](https://innersourcecommons.slack.com/archives/CARTU4XV2/p1581692403087600).

- [x] Check that all titles of the uploads went fine and every video that should be there is there

- [x] Add descriptions to the videos
- [x] Add tags to the videos
- [x] Check if the Playlists are correct and in the correct order. 
- [x]  Check the "this is a authoritative playlist" button on all of them (Set as official series for this playlist)
- [x] Set publishing times on the videos. Probably every video of a series should appear at the same time. This likely triggers the "XYZ uploaded new content" for the people who have subscribed to us "with the bell on" / full upload notification as they say on the videos
- [x] Add links to the articles to the video descriptions
- [x] Ensure that all videos are set to "embeddable"
  - Special checkbox is known to enough people and appears to default to on.
- [x] Ensure that proper license markers as in CC-BY-SA 4.0 are present
- [x] Settings for audience maturity (none) and languages (audio, general: en) are properly set
- [x] Select nice thumbnails for the videos. If the proposed ones are bad one can upload an image (e.g. a video still) for the videos
  - @RJPlog created great suggestions and material here. Splitting this out to its own issue: #243 
- [x] Ping @Tom Sadler (BBC) once there are embeddable links present such that he can update his pull request
  - Only required action here is to update the YT urls to the new non-O'Reilly-URLs
- [x] Come up with Marketing Plans and Publish the results to the respective channels after we ensured that everything is as we want it -> See #60.  Happy to do something there.

rrrutledge:
Before we can work on this we need to finish https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/238 (get the new videos)

rrrutledge:
Some great thoughts from @RJPlog to keep in mind:
* [explanation](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/8cbb02beb41c145ea6bee5235de5c0c206fa8533/assets/thumbnails/thumbnail.asciidoc)
* [example](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/8cbb02beb41c145ea6bee5235de5c0c206fa8533/assets/thumbnails/example5.jpg)

RJPlog:
[...]

PR #241, opened 2020-04-15
german translation of chapter 01 ready for review

RJPlog:
Hi, I translated chapter 01 into german, could you please review?  Any comments are welcome.

Hint:
 I did not change links inside to other chapters, if you like my Translation I will continue and update the links with every new german Translation.


RJPlog:
Hi, meanwhile I worked on issue #214 about the thumbnails for the videos. I generated some examples and doku, than put it at ../assets/thumbnails. Obviou.-sly this new commit was added to this pull request, which I found while trying to generate a new one  :-).

rrrutledge:
I like the research on thumbnails and example and [linked them](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/214#issuecomment-616553809) from https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/214.  They can probably be removed from this pull request, though, so that it can focus on just the German translations.  Since I've linked to your research we won't forget it and can take it forward.

RJPlog:
-Ok, I now cleaned up again all the new commits so that here only the Translation of chapter one should be remaining as before (at least I hope so). Sorry for the confusion.

Issue #243, opened 2020-04-23
Create thumbnails and thumbnails template / documentation for our YT videos


lenucksi:
## Rationale + explanation

YouTube offers us to select a visually appealing thumbnail (a "frame") for our videos. 
@RJPlog has presented some great thoughts and template material for them:
* [explanation](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/8cbb02beb41c145ea6bee5235de5c0c206fa8533/assets/thumbnails/thumbnail.asciidoc)
* [example](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/8cbb02beb41c145ea6bee5235de5c0c206fa8533/assets/thumbnails/example5.jpg)

## ToDo
- [x] Create PR for the template material and explanation @RJPlog 
- [x] Actually create the frames
- [x] Apply the frames to the respective videos.
  - [x] Introduction
  - [x] Trusted Comitter
  - [x] Contributor
  - [x] Product Owner

RJPlog:
I just created a PR (#245) to upload the template. Within the rationale document you can find also an example. Please feel free to add your comments :-).

RJPlog:
I finished  creating all the thumbnails now and applied them to the videos. 
Since we changed the fonts in our template I have to do a final update on the template yet.

Is there a way to mark the done topics in the checkboxes?

lenucksi:
> I finished creating all the thumbnails now and applied them to the videos.
:tada: 

> Is there a way to mark the done topics in the checkboxes?
I've added a few checkboxes for the individual section to the top - you should be able to just check them off. If not, let me know.

rrrutledge:
Sounds like great progress, @RJPlog.

lenucksi:
Anything keeping us from closing this now that #264 is merged too?

rrrutledge:
Looks good to me.

lenucksi:
Looks done to me now.

PR #245, opened 2020-04-30
create thumbnail template (#243) 

RJPlog:
Hi all, 
please find here the propsal for thumbnails with some rationale and a template for that. 

A more detailed description of how to use will follow, first steps are explained directly in the template. Template can be used  with LibreOffice.


RJPlog:
PR replaced by #250

PR #250, opened 2020-04-30
#243 upload template for thumbnail

RJPlog:
I created a template for Thumbnails as in issue #243 described. 
Please find also a rationale and an example.

Don't hesitate to do some comments :-)


assets/thumbnails/thumbnail_howto_use_template.asciidoc
```diff
@@ -0,0 +1,3 @@
+== How to use the template for creating thumbnails
```
rrrutledge, line 1: I think that you could wait to check in this file until you actually have the description.

RJPlog: Thanks for your hint. I removed the file and will check in later when finished. Description in the template should be working for now. :-)

rrrutledge:
🎉

PR #261, opened 2020-05-17
Articles link to isc website

tsadler1988:
Includes a script for substituting article URLs using `urls.yaml`, in case we need to in the future. Had to make a couple of tweaks to urls.yaml to make it work.


rrrutledge:
Amazing, Tom!

PR #278, opened 2020-06-06
CONTRIBUTING.md: Update Translation Instruction

ystk:
I would like to propose to use two-character language code prefix in the file name for translations as follows.
Original: 01-introduction.asciidoc
German translation: de/01-introduction-de.asciidoc
If we use two-character language code in the filename, that makes easier to recognize for other language speaking person when they try to find the files. Also easier to make a regular expression to manipulate it.

$ find . -name '01-intro*' -print
./ja/01-introduction-ja.asciidoc
./de/01-introduction.asciidoc
./01-introduction.asciidoc

The other reason is, using multibyte characters in filename makes difficult to use the filenames from a command line.


rrrutledge:
@ystk I think that folks liked this suggestion.  Would you mind updating the existing articles in _master_ to match this updated scheme (as part of this pull request)?

ystk:
I created a new PR to align this suggestion. Please review #284 .

rrrutledge:
Great!

PR #280, opened 2020-06-09
update the YT urls to the new non-O'Reilly-URLs

RJPlog:
Hi, 
I updated the YT urls to the new non-O'Reilly-URLs in the `urls.yaml`, all thumbnails are confirmed by people pictured, see issue #214 

From what I see this is independent of setting publishing times on the videos, currently both kind of videos are unlisted (the old ones with O'Reilly Branding and the new ones without), which means: 

- Anyone with the video link can see this Video --> links will now lead to the non-O'Reilly videos
- This video won't appear on your channel page. It also won't appear in YouTube search results unless someone adds it to a public playlist.

I will set the publishing times of the playlists separately

### DIFF
--- a/config/urls.yaml
+++ b/config/urls.yaml
@@ -12,7 +12,7 @@
     # URL on the O'Reilly learning plarform.
     oreilly: https://learning.oreilly.com/videos/introduction-to-innersource/9781492041504/9781492041504-video321606
     # URL on YouTube.
-    youtube: https://www.youtube.com/watch?v=tP6K5oF2IxQ
+    youtube: https://www.youtube.com/watch?v=l93ohSHhr5U
   article:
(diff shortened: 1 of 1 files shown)


RJPlog:
thx 😊 

Gesendet von Mail für Windows 10

Von: Tom Sadler
Gesendet: Dienstag, 9. Juni 2020 18:28
An: InnerSourceCommons/InnerSourceLearningPath
Cc: RJPlog; Mention
Betreff: Re: [InnerSourceCommons/InnerSourceLearningPath] update the YT urlsto the new non-O'Reilly-URLs (#280)

@tsadler1988 approved this pull request.
Looks good @RJPlog! I checked the URLs and they all match up.

PR #284, opened 2020-06-18
Filename changes to use two-character language code

ystk:
This PR relate to #278 to rename filenames for translated files to use two-character language code prefix.


ystk:
Oops.
- prefix
+ postfix

Issue #249, opened 2020-04-30
Self-host non-English articles


tsadler1988:
https://github.com/InnerSourceCommons/innersourcecommons.org/pull/159 only self-hosts the English versions of the articles.

rrrutledge:
Exciting!  Thanks for taking this on, @tsadler1988 and @BrittanyIstenes!

rrrutledge:
Is there a way to generate _embed_ URLs that auto-activate a language-specific captioning?  If so, then we can include the videos along with the articles on innersourcecommons.org.

tsadler1988:
> Is there a way to generate _embed_ URLs that auto-activate a language-specific captioning? If so, then we can include the videos along with the articles on innersourcecommons.org.

https://support.google.com/youtube/answer/171780?hl=en-GB

> Make captions automatically appear on an embedded video by adding '&cc_load_policy=1' to the video's embed code.
> 
> You can also choose a caption language for the embedded video. To specify the caption language for the video that you'd like to embed, just add '&cc_lang_pref=fr&cc_load_policy=1' to the video's embed code.
> 
> - 'cc_lang_pref' sets the language for the captions shown in the video.
> - 'cc_load_policy=1' turns captions on by default.
> - 'fr' represents the language code for French. You can look up two-letter language codes in the ISO 639-1 standard.

rrrutledge:
Great, Tom!  So we can include language-specific page of the learning path on _innersourcecommons.org_.

rrrutledge:
Opened https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/306 to adjust the generation script for translated articles and https://github.com/InnerSourceCommons/innersourcecommons.org/pull/206 to display them on innersourcecommons.org

lenucksi:
This is done now, closing.

PR #300, opened 2020-08-20
Generate learning path markdown

tsadler1988:
Adds a script for generating markdown files required for hosting Learning Path on innersourcecommons.org.

Only supports English articles initially - intent is to extend for non-English as part of #249.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -11,3 +11,13 @@ For example, to point all links to innersourcecommons.org:
 ```
 node substitute_article_urls.js isc
 ```
+
+## generate_learning_path_markdown.js
+
+A node script to generate markdown files required for hosting Learning Path on innersourcecommons.org.
+
+### Usage:
+```
+npm ci
+node generate_learning_path_markdown.js
+```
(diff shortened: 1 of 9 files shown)


scripts/README.md
```diff
@@ -10,4 +10,14 @@ node substitute_article_urls.js $TARGET
...
+```
+
+## generate_learning_path_markdown.js
+
+A node script to generate markdown files required for hosting Learning Path on innersourcecommons.org.
+
+### Usage:
+```
+npm ci
+node generate_learning_path_markdown.js
```
lenucksi, line 22: Do we need to integrate this into some sort of GitHub Action?

tsadler1988: That would be the long term aim I think yes. 

In my opinion, we should have a GitHub Action generate the markdown and raise a Pull Request on https://github.com/InnerSourceCommons/innersourcecommons.org if it had changed.

claredillon:
From @lenucksi : Waiting to next week for input from Tom or Russ. Be good to have some desired use documented. This looks like it is related to https://github.com/InnerSourceCommons/innersourcecommons.org/pull/189

tsadler1988:
> Thanks, Tom. This works, and I also wanted to suggest [this improvement](https://github.com/tsadler1988/InnerSourceLearningPath/pull/1/files) that makes it a bit easier to use.

Thanks Russ, cherry picked your commit into this PR.

PR #317, opened 2020-10-01
256 translate trusted committer chapter06

RJPlog:


RJPlog:
puuh, this one was strange with a lot of "contributors". Sounds strange leaving it in english but even more strange to translate it into german....

lenucksi (approved):
Thanks for creating this translation @RJPlog! This looks good and ready to merge to me.

PR #326, opened 2020-10-06
Fixed the broken link of trusted committer

WillemJiang:
The trusted committer pattern link is changed, we need to update the link in the learning path.


rrrutledge (approved):
Thank you!

PR #352, opened 2020-12-01
[Contributor] workbook updates

GoFightNguyen:
In the first segment, question 3 is actually question 2


rrrutledge (approved):
Whoops!  Thank you, Jason!

Issue #362, opened 2021-01-21
Hosting Learning Path on new InnerSource Commons website


tsadler1988:
Assist with https://github.com/InnerSourceCommons/innersourcecommons.net/issues/28

tsadler1988:
Started working on this:
- Scripts: #369 
- New website: https://github.com/InnerSourceCommons/innersourcecommons.net/pull/32

tsadler1988:
Note https://github.com/InnerSourceCommons/innersourcecommons.net/issues/51. 

I assume we won't want to sign off until translations are hosted too? This work has been started under https://github.com/InnerSourceCommons/innersourcecommons.net/pull/52.

tsadler1988:
Remaining task - when selecting a language, we want there to be some feedback for the user. Either:
- Transalte index.md #382
- Add placeholder index.md files to make language selection visible and obvious

tsadler1988:
> Add placeholder index.md files to make language selection visible and obvious

I will do this as part of this issue to complete MVP Learning Path for the new website.

claredillon:
The placeholders have all been edited by @clcoffey to include text to indicate that translations are coming soon.

tsadler1988:
Looks good to me @claredillon. OK to close this issue from my point of view, but I might wait until Thursday's meeting and double check with others.

rrrutledge:
I see translations there - what do you mean about coming soon?

tsadler1988:
> I see translations there - what do you mean about coming soon?

The index pages i.e. the description of each section e.g. https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/382 completed the German index pages translations. On reflection this might be misleading - people may think this means nothing is translated. I believe @RJPlog found this confusing too.

We could change the text, or push to get index page translations ASAP.

clcoffey:
Hi @tsadler1988 and @rrrutledge I tried adding a larger piece of text there that said something along the lines of: This is the index page for the <language> translation of the Learning Path. The index page is in the process of being translated. Press the "Learn" Button below to access the articles which have already been translated.

However, when I added this text the original English text disappeared from the index page. The only way I managed to keep both was by putting all the text in the description: section at the top of each page. In this case, there was no way I could find to format it differently and separate it from the untranslated English text. I would love some suggestions on how we could make it better while we are waiting for translations. Tom if you manage to change one of the pages and be happy with it I can help you with changing the rest.

[...]

Issue #349, opened 2020-11-20
Automate publishing to innersourcecommons.org


rrrutledge:
Every time there is a new article or a new contributor we need to manually run [these steps](https://github.com/InnerSourceCommons/InnerSourceLearningPath/tree/master/scripts#how-to-update-innersourcecommonsorg-with-new-articles) to publish on innersourcecommons.org.

We should automate this process so that it happens automatically every time there is a change to an article in the learning path.  Likely this should be done with GitHub actions.

tsadler1988:
I think we did discussed this in an LP meeting, but we probably want to prioritise innersourcecommons.net. But it might then be easy to also do this for innersourcecommons.org.

rrrutledge:
Yes, that makes sense.

tsadler1988:
New site is live on innersourcecommons.org - previous references to innersourcecommons.net in this issue are obsolete.

tsadler1988:
Started working on this in a fork: https://github.com/tsadler1988/InnerSourceLearningPath/actions/workflows/publish-to-website.yml.

tsadler1988:
PRs opened for a first iteration:
- https://github.com/InnerSourceCommons/innersourcecommons.org/pull/130
- https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/426

Future improvements:
- Trigger on pushes to this repo instead of hourly (discussed in https://github.com/InnerSourceCommons/innersourcecommons.org/pull/130)
- Remove manual step of adding new languages to the website [config.yaml](https://github.com/InnerSourceCommons/innersourcecommons.net/blob/master/config.yaml)

tsadler1988:
Changed the approach to have an Action in the Learning Path which builds the files and commits to a branch of innersourcecommons.org (#432) with a second action on innersourcecommons.org to raise the PR (https://github.com/InnerSourceCommons/innersourcecommons.org/pull/130).

Once #432 is merged and seen to be working, I would consider this issue ready to close.

Remaining manual steps we may want to address (both only occur when adding a completely new language):
- adding new languages to the website [config.yaml](https://github.com/InnerSourceCommons/innersourcecommons.net/blob/master/config.yaml)
- top-level _index pages ([example](https://github.com/InnerSourceCommons/innersourcecommons.org/pull/137))

Issue #248, opened 2020-04-30
Stop self-hosting via raw GitHub


tsadler1988:
In InnerSourceCommons/innersourcecommons.org#159 we are using raw GitHub where we should be using something CDN backed.

lenucksi:
We are actually now hosting with compiled markdown via the new website using GitHub pages instead of always fetching fresh from GitHub.

PR #369, opened 2021-02-04
New website scripts

tsadler1988:
First pass at scripts for #362. 

Before merging, I think we should:
- Document new site generation steps
- Refactor this and the old script to de-duplicate common code
- Use path.join everywhere for cross-platform compatibility

See also https://github.com/InnerSourceCommons/innersourcecommons.net/pull/32, where I've raised rendering issues that need addressing.


rrrutledge:
Wow this is cool, Tom!  I will try to take a look sometime early next week.

tsadler1988:
Tested the refactor with https://github.com/InnerSourceCommons/innersourcecommons.org/pull/252

tsadler1988:
@rrrutledge this is now ready for review. I've addressed everything except GH Actions and translations, which I think can be tackled separately.

tsadler1988:
I had a bit of momentum on this, so have progressed on translations for the new website too (see https://github.com/InnerSourceCommons/innersourcecommons.net/pull/52).

rrrutledge:
Awesome, Tom!  I will try for a review this week.

scripts/README.md
```diff
@@ -15,19 +15,20 @@ For example, to point all links to innersourcecommons.org:
 node substitute_article_urls.js isc
 ```
 
-## generate_learning_path_markdown.js
+## generate_old_site_learning_path_markdown.js, generate_new_site_learning_path_markdown.js
```
rrrutledge, line 18: For future compatibility probably good to label with something besides `old` and `new` (`jekyll` and `hugo`?).  I've painted myself into a corner before with terminology like this one once the `new` website becomes the old one later on.

scripts/README.md
```diff
@@ -38,17 +39,35 @@ Anytime there are _new_ articles written, you need to run the script that genera
...
 3. `cp -r learningpath/* <path-to-innersourcecommons.org-repo>/resources/learningpath/`.
 3. Open a pull request with the modified files in the [InnerSourceCommons/innersourcecommons.org] repo.
 
 Note that these steps only needs to happen when there are new articles written.
 Changes to existing articles will automatically show up on the [innersourcecommons.org] site.
 
-[innersourcecommons.org]: http://www.innersourcecommons.org/
+## How to update innersourcecommons.net
+
+Anytime there are _any changes_, you need to run the script that generates the updated pages on the [innersourcecommons.net] site.
```
rrrutledge, line 55: This works, but just curious why it didn't work out to pull the article text dynamically from the _GitHub_ static content servers?

scripts/section_data.json
```diff
@@ -0,0 +1,34 @@
+[
+  {
+    "learning_path_group": "Introduction",
+    "dirName": "introduction",
+    "workbook": "01-introduction.asciidoc",
+    "translations": ["de", "it", "ja", "zh", "ru"],
+    "image": "images/learn/LP_thumbnail_introduction.jpg",
```
rrrutledge, line 7: Are these files in the other repo?  I feel like they should be here so as to remove this coupling.

scripts/write_markdown_file.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs')
+const YAML = require('yaml')
+const { EOL } = require('os')
+
+module.exports = writeMarkdownFile = (filePath, frontMatter, body) => {
+  const frontMatterTerminator = '---'
+  const originStatement = '<!--- This file autogenerated from https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/scripts/generate_learning_path_markdown.js -->'
```
rrrutledge, line 7: Could update script name here in this comment.

scripts/generate_learning_path_markdown.js
```diff
@@ -1,127 +1,57 @@
...
+    const { dirName, translations } = section
+    const baseReadPath = join('..', dirName)
+    const baseWritePath = join('.', writeDir, dirName)
     mkdirSync(baseWritePath)
 
     translations.concat('' /* The English original */).forEach(async (translation) => {
       const isTranslation = translation !== ''
       const writePath = join(baseWritePath, translation)
-      mkdirSync(writePath)
+      if (createTranslationFolder) mkdirSync(writePath)
```
rrrutledge, line 20: I don't understand why this folder is sometimes needed and sometimes not?

rrrutledge (commented):
This is really good, Tom.  Thanks for this work.  Had a few questions.

scripts/README.md
tsadler1988: Good call, jekyll/hugo make more sense.

scripts/README.md
tsadler1988: I followed the implementation proposed here - https://github.com/InnerSourceCommons/innersourcecommons.net/issues/28 - which does it at build time, rather than run time, and it didn't occur to me to challenge it. This does make #349 a lot more important, but I think that's OK as we wanted to do this anyway.

I think this is probably better for a few reasons:
- Less custom JavaScript added to innersourcecommons.net, making development easier for future contributors to innersourcecommons.net
- More scaleable - don't need to rely on raw GitHub, which isn't really designed for static hosting as it isn't CDN fronted - see #248
- Marginally more performant for the user - don't need to wait for the additional network request to raw GItHub

Only downside is the need to complete #349, preferably before innersourceommons.net goes live!

scripts/section_data.json
tsadler1988: Most of the time the thumbnail is retrieved from YouTube (see https://github.com/InnerSourceCommons/innersourcecommons.net/blob/master/content/learn/learning-path/contributor/01.md for example) - this is just a backup for where there is no video, which is only the workbooks.

I do think we need to choose an image for workbooks, as using intro thumbnails isn't really relevant. I tried rendering the site without images for workbooks but it doesn't really work.

Can we raise an issue for a workbook image separately, and address the coupling at the same time?

scripts/write_markdown_file.js
tsadler1988: Will do. I will link to the folder, as there are 2 scripts that generate markdown now.

scripts/generate_learning_path_markdown.js
tsadler1988: innersourcecommons.net doesn't use separate folders for translations. They appear in the same folder but with a special suffix (see https://github.com/InnerSourceCommons/innersourcecommons.net/pull/52).

rrrutledge (approved):
Looks good thank you!

scripts/section_data.json
tsadler1988: https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/375

rrrutledge: It seems like the way to resolve the coupling issue would be to move the images to the InnerSourceLearningPath repo and then copy them over to innersourceommmons.org along with the articles - no?

PR #380, opened 2021-03-09
List Co-Committers as Learning Path Contributors

rrrutledge:
* When someone leaves a commit suggestion that is merged, they show up as a 'co-author' on that commit
* GitHub exposes this information in their graph API (https://bit.ly/3cdyYQQ)
* You can see the script output at https://github.com/InnerSourceCommons/innersourcecommons.org/pull/257


tsadler1988 (approved):
Nice @rrrutledge, some ES6 tricks in there I haven't seen before!

rrrutledge:
Thanks for the review!

PR #381, opened 2021-03-10
New website index pages

tsadler1988:
This enables auto-generation of the required index pages for the new website (see https://github.com/InnerSourceCommons/innersourcecommons.net/pull/52).

If there is no index page data for a given translation, it will fall back to English, but we should still reach out to translators to get language specific index pages.


scripts/generate_learning_path_markdown.js
```diff
@@ -1,14 +1,18 @@
...
+
 const getContributors = require('./get_contributors')
 const mkdirSync = require('./mkdir_sync')
 const getArticleFiles = require('./get_article_files')
+const writeMarkdownFile = require('./write_markdown_file')
 
 const sections = require('./section_data.json')
 
-module.exports = async (writeDir, generatorFn, workbookFn, createTranslationFolder) => {
+module.exports = async (writeDir, generatorFn, workbookFn, isJekyllSite) => {
```
rrrutledge, line 12: Better variable naming/purpose so as not to bleed implementation details out of this function

rrrutledge (approved):

Issue #382, opened 2021-03-11
Translate index.md


tsadler1988:
The [new website](https://innersourcecommons.net/learn/learning-path/) requires [index.md](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/introduction/index.md) files for each language and each section. 

https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/362 will add English placeholders, but ideally we want them translated so that users can navigate the Learning Path in their chosen language.

Introduction
- [X] Chinese
- [X] Japanese
- [x] Italian
- [x] Spanish
- [X] German
- [X] Russian
- [X] French

Trusted Committer
- [X] Chinese
- [X] Japanese
- [X] German

Contributor
- [X] Chinese
- [X] Japanese
- [x] Italian
- [X] German

Product Owner
- [X] Chinese
- [X] Japanese
- [X] German

tsadler1988:
- We need to progress #349 so that these get added to the website automatically.
- We also need to add the placeholders created by https://github.com/InnerSourceCommons/innersourcecommons.org/pull/97 to this repo, so we don't lose the placeholder text

tsadler1988:
@ystk @WillemJiang @OpenPj @dottorblaster @dicortazar @adadai88 @niminghong @tanzhongyi003 @yvonneT1 - reaching out to our wonderful translation contributors, apologies if I've missed anyone! 

If you have the time, please could you look at translating the index.md files in this repository. The text in the index.md files is used for the short description of each section on the website. As an example, #385 did this for German). We've added some placeholder text for now as you can see below in the below screenshot of the [Japanese Learning Path homepage](https://innersourcecommons.org/ja/learn/learning-path/).

![image](https://user-images.githubusercontent.com/5602024/113182998-6f279680-924b-11eb-844c-b4f758b61c50.png)

OpenPj:
Probably the next week I'll find some time to work on this 👍

rrrutledge:
Hi, @OpenPj, did you still want to try for this?

OpenPj:
Sure, sorry for this delay, I'll try to work on this today!

OpenPj:
Just created a new pull request for this, anyway it includes only the current sections translated in italian:
https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/412
Hope to find some time during these days to finish the overall translation.

tsadler1988:
[...]

PR #385, opened 2021-03-22
#382 creation of german index.md files

RJPlog:
Hi Tom, I created german translations of the md files as you suggested for each section. I put it into the ../de folder. 

In the introduction file you have a variable featured: true, I just kept that, even if not knowing what it means and why it is not in the other sections.

Please let me know if I have some misunderstanding.

@lenucksi maybe you would like to review the translation.

### DIFF
--- a/contributor/de/index.md
+++ b/contributor/de/index.md
@@ -1,0 +1,4 @@
+---
+title: Learning Path - Contributor
+---
+Im Kapitel Contributor wird beschrieben, was es bedeutet ein InnerSource Contributor zu sein, welche Verhaltensmuster Dich zu einem erfolgreichen Contributor machen, welche Mechanismen es gibt, um Beiträge zu erstellen und welche Vorteile sich durch InnerSource für Dein Team und Deine Organisation ergeben.
--- a/introduction/de/index.md
+++ b/introduction/de/index.md
@@ -1,0 +1,5 @@
+---
+title: Einführung
+featured: true
+---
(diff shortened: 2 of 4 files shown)


lenucksi:
Excellent @RJPlog, thanks a lot. I'll happily review them and get them merged. Might even happen today evening.

lenucksi (approved):
These changes look are crisp and easy to understand. Also, I could not find any mistakes with regard to spelling or grammar. 
This looks good to merge to me.

@tsadler1988 Do you want to give the mentioned front matter variable a look and maybe add needed changes as commit suggestions?

tsadler1988:
Thanks for jumping on this so quickly @RJPlog!

Do we want to translate the titles, or all these all terms that we don't translate?

tsadler1988:
> In the introduction file you have a variable featured: true, I just kept that, even if not knowing what it means and why it is not in the other sections.

Yes this is correct, it means the Introduction section will appear larger on this page:

image

RJPlog:
Thanks for explaining the matter with the 'featured', understood. Concerning the translation of the headings, we once said we will not translate keywords link learning path/trusted committer..... So in oder to follow the translations in the sections I also left it as it is, only with introduction I felt it would look better if this one will be translated into german. Means, this was done on purpose, but I  am happy to change if you suggest to do so

tsadler1988:
I thought that might be the case, thanks for clarifying @RJPlog. Approved 😄 .

tsadler1988 (approved):

rrrutledge:
Great work everyone!

Issue #294, opened 2020-08-06
"Getting Started with InnerSource" segment outline


rrrutledge:
There is interest in making a single segment that outlines how to get started in InnerSource.  We can start this work by creating a segment outline.

rrrutledge:
Here is a simple InnerSource checklist file that we've shared.  We haven't given ours much attention for a while, so it is definitely missing things.

https://docs.google.com/document/d/1-9ea5OlkgRVTECa-4NTzthpmM4KwUMdyVvR1Ep9jSm4/edit

markdav-is:
I think this is a great idea!

markdav-is:
that link https://docs.google.com/document/d/1-9ea5OlkgRVTECa-4NTzthpmM4KwUMdyVvR1Ep9jSm4/edit gives me access denied error.  can you share an open link?

rrrutledge:
Sorry, just saw this message.  Try again!

rrrutledge:
Hi, @tapjdey, just checking are you still interested in working on this?

tapjdey:
Hi @rrrutledge , sorry I've been a bit busy this month. I'll create the PR soon.

rrrutledge:
Ah, nice!  Sounds good.

tapjdey:
@rrrutledge Sorry I haven't been able to work on this. I'll get an update on the first week on June

markdav-is:
same here.  I just started a new job and drowning in onboarding there.  might be a few months before I come up for air.

rrrutledge:
No problem, folks.  Thanks for the update.  Good luck on your stuff!

PR #386, opened 2021-03-25
Create introduction-outline.md

rrrutledge:
* Placeholder for an outline giving in introduction to the InnerSource Commons.
* @tapjdey can do a pull request against this template when working on [Getting Started with InnerSource segment outline](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/294).
* The thing to review here is do we like creating a new section in the learning path called "InnerSource Commons"?

### DIFF
--- a/innersourcecommons/introduction-outline.md
+++ b/innersourcecommons/introduction-outline.md
@@ -1,0 +1,24 @@
+# Audience
+
+_Who is the intended audience for your segement?_
+_Be as specific as possible, as it will help you to focus and streamline your content._
+
+# Takeaways
+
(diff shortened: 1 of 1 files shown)


lenucksi (approved):
Looks good to me.

and...
* I 
* like
* the way
* of 
* making 
* the 
* style and purpose of the outline clear.

lenucksi:
Might have merged a bit bit prematurely. This is a good template to start new sections with though. I would have though about moving it to some part of meta, but in the end that's a fair new section. We could possible add a section to the README.md indicating the separation of the original LP and additional modules that are differently named.

rrrutledge:
Thanks!

Issue #254, opened 2020-05-09
Translate "Contributor" articles to German


rrrutledge:
It's great to have the "Introduction" segment of the "Trusted Committer" section [translated in German](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/trusted-committer/de/01-introduction.asciidoc).  Let's do the same for the articles in the ["Contributor" section](https://github.com/InnerSourceCommons/InnerSourceLearningPath/tree/master/contributor).

Similar to "Trusted Committer", translations should be located in a `de` folder.

- [x] introduction

- [x] becoming a contributor

- [x] contributor ethos

- [ ] mechanics of contributing

- [ ] benefits of contribution

- [ ] conclusion

RJPlog:
Hi all, I would like to start with this issue. Any objections?

lenucksi:
This sounds great to me, you'd be most welcome to do so from my point of view. I'll see into reviewing the results once I find a bit of time for that. Maybe @arnom-ms or @MaineC might find a bit of time for that too.

rrrutledge:
Yes, please do, @RJPlog !!

rrrutledge:
Looking forward to seeing them live on _innersourcecommons.org_.

tsadler1988:
Moving back to TODO pending further articles.

rrrutledge:
May be some time to work on this after the holiday season (early-to-mid August).

PR #387, opened 2021-03-28
#254 german translation contributor chapter 03

RJPlog:
Hi, with this chapter I had some difficulties doing a strict literal translation. So I took a little bit more freedom, at least for me that fits better. Please have a look what you are thinking about, I am looking forward to see your comments.


lenucksi (approved):
I read it, is covers all the points from the English script I remember, was easy to understand and I didn't find any obvious grammar or spelling issues. Hence: Looks good to me.

@arnom-ms Can you take a second look regarding grammar and spelling issues I might have missed? And of course other points you might find.

lenucksi:
Let's merge this as is, I think it looks good to me. If any more errors are found, we can still fix them in a new PR.

PR #390, opened 2021-03-31
Fix new site links

tsadler1988:
Links in articles are currently broken on the new site. This PR fixes this, and updates the `urls.yaml` to reflect that the new site is live on innersourcecommons.org.


rrrutledge (approved):
Great!  Thank you!

PR #391, opened 2021-03-31
Clean up scripts for old website

tsadler1988:
~Note this has been branched from #390, so also contains those changes.~ #390 merged.


lenucksi:
I remember there was something about tracing possible links in our videos that we'd need to keep active. Do we need to address this here too?

tsadler1988:
> I remember there was something about tracing possible links in our videos that we'd need to keep active. Do we need to address this here too?

I'm not sure about this - do you mean links referenced in the videos? If so, that would probably need a redirect/permalink set up in the new Hugo site. Happy to take a look, but would prefer raising as a separate issue.

rrrutledge:
Why does it appear that we are making functional code changes, here?  Is it just refactoring?

lenucksi:
> > I remember there was something about tracing possible links in our videos that we'd need to keep active. Do we need to address this here too?
> 
> I'm not sure about this - do you mean links referenced in the videos? If so, that would probably need a redirect/permalink set up in the new Hugo site. Happy to take a look, but would prefer raising as a separate issue.

Created #393 for this.

tsadler1988:
> Why does it appear that we are making functional code changes, here? Is it just refactoring?

Should just be a refactor, but it is quite a big one. I've pulled logic out of `generate_hugo_learning_path_markdown.js` and put it into `generate_learning_path_markdown.js`, which might be why it looks like functional change.

rrrutledge:
I'd like to review by actually trying this out but not sure exactly when I'll have time.

tsadler1988:
No worries @rrrutledge, I don't think this is blocking anything right now so no rush.

rrrutledge (approved):
Looks great!  I used it for https://github.com/InnerSourceCommons/innersourcecommons.org/pull/115

lenucksi:
Sounds like this could be ready to be used in GitHub actions for automation purposes?

rrrutledge:
> Sounds like this could be ready to be used in GitHub actions for automation purposes?

I think so - just need to wire it up.

PR #405, opened 2021-05-25
Fix broken links to web site

johtani:
Fix broken links to learning path on web site


lenucksi (approved):
Excellent, thanks for finding and fixing these @johtani!

PR #415, opened 2021-06-11
Create 06-conclusion.asciidoc

fioddor:
Review warning: This is my first contribution to ISC,
According to https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/efcca61a21414ec676c1eb14efa4945aa2b7806e/CONTRIBUTING.md, _the filename of the article should be changed to include two-character language code prefix_, but I see that **es** is keeping the original name under the `es/` folder, so I've folowed the _de facto_ standard.


rrrutledge:
Great work getting this started!

PR #420, opened 2021-06-17
Update 01-introduction.asciidoc

SJVle:
Punctuation,  do we let the "Traduit avec www.DeepL.com/Translator (version gratuite)" ?

### DIFF
--- a/introduction/fr/01-introduction.asciidoc
+++ b/introduction/fr/01-introduction.asciidoc
@@ -1,13 +1,13 @@
 == Introduction
 
-Ce parcours d'apprentissage est une introduction à l'InnerSource
+Ce parcours d'apprentissage est une introduction à l'InnerSource.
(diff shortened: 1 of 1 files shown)


lenucksi:
> Punctuation, do we let the "Traduit avec [www.DeepL.com/Translator](http://www.DeepL.com/Translator) (version gratuite)" ?

Thanks for creating this. I'd like to remove the DeepL line, but I'm not 100% if we need to state any attribution to using their service. Do we have people with opinions on that? I couldn't find anything indicating need to attribution myself.

lenucksi (approved):
Language-Change wise this looks good to me. So approval from me.

rrrutledge:
> I'd like to remove the DeepL line, but I'm not 100% if we need to state any attribution to using their service. Do we have people with opinions on that? I couldn't find anything indicating need to attribution myself.

Discussed some reasons on the call to leave it in.

lenucksi:
Thanks again for creating this @SJVle :tada: :smile: 
Merging this in!

PR #431, opened 2021-06-30
Updating youtube link

clcoffey:
Updating the Youtube Link for the Contributor content with the new youtube link of the modified video.

Related to https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/122

PR #432, opened 2021-07-06
Learning path automation

### LINKED ISSUE #349: Automate publishing to innersourcecommons.org

tsadler1988:
Along with https://github.com/InnerSourceCommons/innersourcecommons.org/pull/130, this fixes #349.

This new GitHub Action will (if relevant files have changed on master):
- Checkout LearningPath and innersourcecommons.org repos
- Checkout or create a `publish-latest-learning-path` branch of innersourcecommons.org
- Run the scripts to generate website markdown from learning path articles and data
- Commit and push to the `publish-latest-learning-path` branch of innersourcecommons.org

Opening a Pull Request will be done a separate action run on innersourcecommons.org repo, due to limitations of GitHub deploy keys (see https://github.com/InnerSourceCommons/innersourcecommons.org/pull/130).

This has been tested against earlier commits, where the action was configured to build from this branch - https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/workflows/publish-to-website.yml - and the output used to open a PR updating the website - https://github.com/InnerSourceCommons/innersourcecommons.org/pull/134.

### DIFF
--- a/.github/workflows/publish-to-website.yml
+++ b/.github/workflows/publish-to-website.yml
@@ -6,7 +6,7 @@ env:
 on:
   push:
     branches:
-      - learning-path-automation
+      - master
     paths:
       - .github/workflows/publish-to-website.yml
       - config/urls.yaml
--- a/.github/workflows/publish-to-website.yml
+++ b/.github/workflows/publish-to-website.yml
@@ -30,7 +30,7 @@ jobs:
         with:
           ssh-key: ${{ secrets.WEBSITE_DEPLOY_KEY }}
           path: ./innersourcecommons.org
-          repository: tsadler1988/innersourcecommons.org
+          repository: InnerSourceCommons/innersourcecommons.org
           fetch-depth: 0
       - name: Branch innersourcecommons.org
         working-directory: innersourcecommons.org
--- a/.github/workflows/publish-to-website.yml
+++ b/.github/workflows/publish-to-website.yml
@@ -1,0 +1,56 @@
+name: Publish to website
+
+env:
+  BRANCH_NAME: publish-latest-learning-path
+
+on:
+  push:
+    branches:
+      - learning-path-automation
+    paths:
+      - .github/workflows/publish-to-website.yml
+      - config/urls.yaml
+      - contributor/**
+      - introduction/**
+      - product-owner/**
+      - scripts/**
+      - trusted-committer/**
+      - workbook/**
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Learning Path
+        uses: actions/checkout@v2
+        with:
+          path: ./InnerSourceLearningPath
+      - name: Checkout innersourcecommons.org
+        uses: actions/checkout@v2
+        with:
+          ssh-key: ${{ secrets.WEBSITE_DEPLOY_KEY }}
+          path: ./innersourcecommons.org
+          repository: tsadler1988/innersourcecommons.org
+          fetch-depth: 0
+      - name: Branch innersourcecommons.org
+        working-directory: innersourcecommons.org
+        run: |
+            git checkout --track origin/$BRANCH_NAME || git checkout -b $BRANCH_NAME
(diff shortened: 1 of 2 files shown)


tsadler1988:
@lenucksi / @rrrutledge would either or both of you be able to review this please?

config/urls.yaml
```diff
@@ -226,7 +226,7 @@
   video:
     isc: https://innersourcecommons.org/learn/learning-path/contributor/02/
     oreilly: https://learning.oreilly.com/learning-paths/learning-path-the/0636920338833/0636920338802-video329500
-    youtube: https://youtu.be/S0Gps2AbZ7M
+    youtube: https://www.youtube.com/watch?v=S0Gps2AbZ7M
```
rrrutledge, line 229: Just curious why this change?

.github/workflows/publish-to-website.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Learning Path
+        uses: actions/checkout@v2
+        with:
+          path: ./InnerSourceLearningPath
+      - name: Checkout innersourcecommons.org
+        uses: actions/checkout@v2
+        with:
+          ssh-key: ${{ secrets.WEBSITE_DEPLOY_KEY }}
```
rrrutledge, line 31: This is entered in by repo admins somewhere in the UI?

.github/workflows/publish-to-website.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: Build
+        working-directory: InnerSourceLearningPath/scripts
+        run: |
+            npm ci
+            TOKEN=${{ secrets.GITHUB_TOKEN }} node generate_learning_path_markdown.js
+            cp -r newsite/. ../../innersourcecommons.org/content/learn/learning-path/
+      - name : Commit
```
rrrutledge, line 49: Is there a way to open a PR or do we still have to do that manually?

rrrutledge: Maybe we should just list directly to _main_ and skip the whole branch thing.

rrrutledge (approved):
Very neat to see!

config/urls.yaml
tsadler1988: When this was updated in #431 it used the `youtu.be` format, but the script expects `www.youtube.com` format - all other YouTube URLS in this file are in that format. The script needs to extract the video ID from the URL.

.github/workflows/publish-to-website.yml
tsadler1988: Correct. @voborgus added the public key as a Deploy Key  for https://github.com/InnerSourceCommons/innersourcecommons.org, and I added the private key to https://github.com/InnerSourceCommons/InnerSourceLearningPath as an Action secret.

.github/workflows/publish-to-website.yml
tsadler1988: Discussed in https://github.com/InnerSourceCommons/innersourcecommons.org/pull/130, but to summarise:
- Deploy Keys do not have GitHub API access, so this Action cannot raise a PR. Instead, there is another Action on the innersourcecommons.org repo which raises a PR whenever the `publish-latest-learning-path` branch is pushed to (added in https://github.com/InnerSourceCommons/innersourcecommons.org/pull/130). We considered alternatives to using a Deploy Key, but stuck with this solution.
- Original plan was to push directly to main, but this would trigger a website rebuild into production and the preference was to have a manual step to reduce the risk of an outage of the website.

Issue #428, opened 2021-06-21
Automation to ensure links in translations are correct


tsadler1988:
We are sometimes getting situations where links in translated articles are broken (see https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/419).

There may be a way to ensure correctness through automation, using the existing [scripts](https://github.com/InnerSourceCommons/InnerSourceLearningPath/tree/master/scripts) and [urls.yaml](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/master/config/urls.yaml).

tsadler1988:
Work in progress: https://github.com/InnerSourceCommons/InnerSourceLearningPath/tree/check-translation-links

PR #419, opened 2021-06-14
Fix broken links for translated articles

ystk:
This PR fixes links for Chinese, Russian and Spanish articles.
Currently all links in Chinese articles point to Chinese translation, but the other language translations are not. I think it depends on the status of translation for each language.

PR #427, opened 2021-06-21
Format and link

SJVle:
Last (?) corrections : format, links,and so on.

PR #434, opened 2021-07-10
Modified 01 links to point landing pages

ystk:
This PR replace all 01 links to landing page for each segment. (Discussions were made in #419).
This PR also includes changes to point translated articles for Chinese and Japanese articles.
These changes were made by using sed command. I have tested the changes for one article in each segment only. But it should be okay because the other changes use exactly the same rule.

I hope these changes will pass the test in #428 .


ystk:
I also have another changes for filename fixes. 
If you are okay, I would like to include the filename changes in this PR. 
You can see the changes at the following URL.
https://github.com/ystk/InnerSourceLearningPath/compare/landing-page-link...ystk:rename-filenames-for-translations

rrrutledge:
Thank you, @ystk!  Will take a look at it probably in the next few days.  Have been on vacation.

tsadler1988 (approved):
Looks good to me, thanks @ystk! Manually checked a selection of the links and all were correct. #428 hasn't been picked up yet, so just manually reviewing and checking for now 😄 

I see the file name changes haven't been included, assume you'll make a separate PR for that?

ystk:
> I see the file name changes haven't been included, assume you'll make a separate PR for that?

Yes, I will. Before making the changes, I try to resolve merge conflicts on #427 .

PR #443, opened 2021-08-07
Minor fixes for Japanese translations

ystk:
This PR fixes the following points.
- Fix spelling inconsistency for Contributor (Changed 'コントリビュータ' to 'コントリビューター')
- Fix spelling inconsistency for InnerSource Commons (Changed 'InnerSource Commons' to "InnerSourceコモンズ")
- Fix special term Trusted Committer (Changed 'トラステッドコミッター' to 'Trusted Committer')
- Omit unnecessary whitespaces

PR #444, opened 2021-08-07
Changed links to point to proper content

ystk:
This PR fixes a bug that was introduced in #434 .
In the changed sentence, the link should point to introduction article in trusted committer segment. Because the sentence mentioned a video which is available in the article.
The same change included in #443 for Japanese article.


lenucksi (approved):
Thanks @ystk for catching this and fixing it! :tada:

PR #446, opened 2021-08-20
Clarify Contributor workbook question

rrrutledge:
Fixes https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/122

### DIFF
--- a/workbook/04-contributor.asciidoc
+++ b/workbook/04-contributor.asciidoc
@@ -30,18 +30,19 @@ Why 4 is correct: Contributions go beyond just code. There are many aspects that
 
 ===== Question 2.  What is a productive assumption for a contributor to make when joining an InnerSource project?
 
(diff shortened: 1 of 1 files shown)


lenucksi (approved):
I like those clarifications and they LGTM but would like to hint to another addition @Ludmila-N made in https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/122#issuecomment-830698624: 
"Question 1, answer 4 has the explanation that was never mentioned in the video, suggest rewriting this option:" (Contributor section, Contributor Ethos)

Maybe we can get another commit on those before we merge it? Maybe we can rephrase it along the lines of "a proven pattern for this problem ist ... 30-days-of-warranty" or so?

rrrutledge:
> Question 1, answer 4 has the explanation that was never mentioned in the video

Although not in the video, it is in the article, which we've said before is fair game.

PR #457, opened 2021-11-15
Add script to check links in translations

### LINKED ISSUE #428: Automation to ensure links in translations are correct

tsadler1988:
Fixes #428.

This script modifies links in translated articles in 2 cases:
- If the linked article has also been translated, make sure that is linked to
- If the linked article has not been translated, make sure the default (English) article is linked to

I have run the script against the current state of the Learning Path articles, which has uncovered some examples of incorrect links.

**What next?**
Assuming this gets merged:
- Run this in a GitHub Action 
- Update `substitute_article_urls.js` to work with translations
- Look at these scripts holistically for refactoring opportunities (e.g. multiple scripts need to retrieve article files)


tsadler1988:
@rrrutledge - I know you're familiar with the scripts we use in the Learning Path, would you be able to review this please?

scripts/substitute_article_urls.js
```diff
@@ -1,3 +1,5 @@
+// TODO: Update to work with translations; may require changing the urls.yaml
```
rrrutledge, line 1: Are we even using the functionality in this file anymore?  It looks like we're just always hard-linking to innersourcecommons.org.

rrrutledge (approved):
This is great!  I'll admit that when I first heard about the idea for this work I questioned how/if it could be done reasonably.  Great work!

rrrutledge:
As part of this work or in another issue we could also ensure that we use translated titles when linking to translated articles.

scripts/substitute_article_urls.js
tsadler1988: Good point, I can't think of any requirements we have that this would help, unless we're planning to upload to other platforms or re-uploading to O'Reilly.

rrrutledge: Not that I know of.

PR #477, opened 2022-07-26
Move French Trusted Committer into correct folder structure

tsadler1988:


rrrutledge (approved):
Thanks for doing this, Tom

PR #479, opened 2022-08-03
Create 03-keeping-the-community-healthy-fr.asciidoc

nysenthil:
Machine translated french text for this article that needs to be verified by a french speaker.


nysenthil:
I reviewed it with French speaking IBM colleague and merging it now.

PR #506, opened 2022-11-02
Ran the link verification script.

nysenthil:
We modified seven files to fix the language related part of the URLs in the translated articles.


rrrutledge (approved):
Great work @nysenthil ❗

Issue #503, opened 2022-11-01
Build failing (rate limited)


rrrutledge:
We can't publish to the website.  See https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/runs/3228272591/jobs/5284148344

rrrutledge:
@marshmallowrobot has some more detail.

rrrutledge:
Some ideas:
* Can we cache results when fetching user info?
* Can we batch calls together (rather than one request per file)?

marshmallowrobot:
Adding some more details...

The build fails in the `get_contributors.js` file with the following error messages:
```
(node:1722) UnhandledPromiseRejectionWarning: HttpError: You have exceeded a secondary rate limit. Please wait a few minutes before you try again.
[109](https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/runs/3228272591/jobs/5284148344#step:6:110)
    at /home/runner/work/InnerSourceLearningPath/InnerSourceLearningPath/InnerSourceLearningPath/scripts/node_modules/@octokit/request/dist-node/index.js:66:23
[110](https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/runs/3228272591/jobs/5284148344#step:6:111)
    at runMicrotasks (<anonymous>)
[111](https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/runs/3228272591/jobs/5284148344#step:6:112)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
[112](https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/runs/3228272591/jobs/5284148344#step:6:113)
    at async module.exports (/home/runner/work/InnerSourceLearningPath/InnerSourceLearningPath/InnerSourceLearningPath/scripts/get_contributors.js:13:49)
[113](https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/runs/3228272591/jobs/5284148344#step:6:114)
    at async /home/runner/work/InnerSourceLearningPath/InnerSourceLearningPath/InnerSourceLearningPath/scripts/generate_learning_path_markdown.js:65:30
```


For everyone's reference, here are GitHub's rate limit guidelines:
- Individual calls cannot request more than **500,000 total [nodes](https://docs.github.com/en/graphql/guides/introduction-to-graphql#node)**.
[...]

PR #512, opened 2022-11-15
503 throttle graphql requests

marshmallowrobot:
This PR introduces the throttling plugin to our GraphQL calls. It prevents the [rate-limiting errors](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/503) we were seeing prior to this, but at the cost of increased time. It now takes a couple of minutes for `scripts/generate_learning_path_markdown.js` to complete.

The throttling plugin requires two new functions to be defined: `onRateLimit()` and `onSecondaryRateLimit()`, neither of which should get triggered if the throttling is working as intended. I didn't do anything different from the documentation in those. 

I also made small changes to the auth and query objects. The [@octokit README](https://github.com/octokit/graphql.js#variables) specifically mentions string template literals as an anti-pattern, so I removed them, and in the query, used a variable instead.

### DIFF
--- a/scripts/generate_learning_path_markdown.js
+++ b/scripts/generate_learning_path_markdown.js
@@ -57,7 +57,6 @@ const getYouTubeImage = (youTubeCode) => {
 
       let { indexContent, ...indexFrontMatter } = yamlFront.loadFront(fs.readFileSync(indexReadPath, 'utf-8'), { contentKeyName: 'indexContent' })
       indexFrontMatter.image = section.image
-      indexFrontMatter.contributors = await getContributors(relative('..', readPath))
       indexFrontMatter.weight = sectionDataIndex
 
       if (!indexTranslated) {
--- a/scripts/get_contributors.js
+++ b/scripts/get_contributors.js
@@ -2,17 +2,16 @@ require('dotenv').config()
 const { Octokit } = require("@octokit/core")
 
 module.exports = async function (filepath) {
-  // Translate Windows-style paths to Unix-style paths.
-  filepath = filepath.replace(/\\/g, "/")
+  const unixStyleFilepath = filepath.replace(/\\/g, "/")
   const octokit = new Octokit({
     auth: process.env.TOKEN
   })
   const { repository: { object: { history } } } = await octokit.graphql(
-    `{
+    `query ($filepath: String!) {
       repository(owner: "InnerSourceCommons", name: "InnerSourceLearningPath") {
         object(expression: "main") {
           ... on Commit {
-            history(first: 100, path: "${filepath}") {
+            history(first: 100, path: $filepath) {
               totalCount
               nodes {
                 authors(first: 100) {
@@ -44,7 +43,9 @@ module.exports = async function (filepath) {
           }
         }
       }
-    }`
+    }`, {
+      filepath: unixStyleFilepath
+    }
   )
 
   if (history.totalCount > 100) {
--- a/scripts/get_contributors.js
+++ b/scripts/get_contributors.js
@@ -10,7 +10,7 @@ const graphqlWithAuth = graphql.defaults({
 module.exports = async function (filepath) {
   // Translate Windows-style paths to Unix-style paths.
   filepath = filepath.replace(/\\/g, "/")
-  const { repository: { object: { history }}} = await graphqlWithAuth(
+  const { repository: { object: { history } } } = await graphqlWithAuth(
     `{
       repository(owner: "InnerSourceCommons", name: "InnerSourceLearningPath") {
         object(expression: "main") {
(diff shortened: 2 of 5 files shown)


rrrutledge:
This work is amazing, @marshmallowrobot ❗️ I added @tsadler1988 to this pull request, who is the original author of a lot of this code.  I’m excited to take a look, too!

scripts/generate_learning_path_markdown.js
```diff
@@ -57,7 +57,6 @@ const getYouTubeImage = (youTubeCode) => {
 
       let { indexContent, ...indexFrontMatter } = yamlFront.loadFront(fs.readFileSync(indexReadPath, 'utf-8'), { contentKeyName: 'indexContent' })
       indexFrontMatter.image = section.image
-      indexFrontMatter.contributors = await getContributors(relative('..', readPath))
```
rrrutledge, line 60: Is this contributor fetching being totally removed or is it moved somewhere else somehow?

rrrutledge (commented):
Thank you for this work, @marshmallowrobot ‼️

scripts/generate_learning_path_markdown.js
marshmallowrobot: I removed it. It is not moved elsewhere. I did not see where this information was being displayed on the website. Did I miss something? Should we put this line back in?

scripts/get_contributors.js
```diff
@@ -1,21 +1,33 @@
...
+        if (retryCount <= 1) {
+          octokit.log.info(`Request quota reached for ${options.method} ${options.url}, retrying after ${retryAfter} seconds`)
+          return true
+        }
+        octokit.log.warn(`Request quota exhausted for ${options.method} ${options.url}`)
+        return false
+      },
+      onSecondaryRateLimit: (retryAfter, options, octokit) => {
+        // does not retry, only logs a warning
+        octokit.log.warn(`SecondaryRateLimit hit for request ${options.method} ${options.url}`)
```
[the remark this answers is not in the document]
marshmallowrobot: Yes, execution should stop if we hit the secondary rate limit error. However, we should not hit this error. The throttling prevents it from getting triggered.

Would you rather retry in this case?

scripts/generate_learning_path_markdown.js
marshmallowrobot: Yes, I see the contributors on articles. They are not affected by this change.

There were (previously) 3 calls to `getContributors()`:
 - Line 60 (this one), now removed. This one appears to fetch contributors for paths like `/introduction`, or `/trusted-committer`. Looking at these pages in the website, we don't seem to show contributors. 
 - Line 73 (old) / Line 72 (new). This fetches contributors for each articles. 
 - Line 96 (old) / Line 95 (new). This fetches contributors for the workbook. 

Removing Line 60 only eliminates like 3 extra calls to GitHub's GraphQL. But, since these calls are to parent folders/paths, these are the requests that triggered the *"This script needs updating to handle >100 commits"* warnings in `getContributors()` script. It's a lot of data to fetch and process for info we don't appear to display.

That said, if I've developed contributor blindness and we DO show this info on the website somewhere and I just can't see it, I'm happy to put this line back in.

scripts/get_contributors.js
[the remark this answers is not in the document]
marshmallowrobot: Just now updated the PR to retry onSecondaryRateLimit(). TY!

scripts/generate_learning_path_markdown.js
rrrutledge: That's true we just need contributors on the articles themselves not the index.

scripts/get_contributors.js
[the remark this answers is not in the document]
rrrutledge: Looks great!

rrrutledge (approved):
Great stuff!  So does the build actually finish now?

lenucksi:
> > Great stuff! So does the build actually finish now?
> 
> Locally, YES! It does. Takes 2.5-3 min to finish, but its consistent, and no rate limit responses from GitHub API

Sounds great, how about we try it? 
I still remember a 10 minute pre-commit script, (not CI, just pre-commit) so 2.5-3 minutes sounds good to me. (Not that one should compare themselves to that pre-commit script. Company was fine, effect of the script was also useful, still too long. :wink: )

lenucksi:
> > Sounds great, how about we try it?
> 
> Ok! What do I do to try it? If I merge the PR, does that magically kick something off? Or do I have to go run another script somewhere? What's this process like?

From what I can see here: https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/main/.github/workflows/publish-to-website.yml it should run. However, it appears to only run if certain directories have been touched by a commit. So might need a manual run.

rrrutledge:
Woah!  So amazing?  So the website is updated now?

rrrutledge:
I see it!  I see it!  Wow‼️

PR #522, opened 2023-01-10
375-workbook-images - Use custom generated images for workbooks

marshmallowrobot:
Hey ya'll,

This PR introduces puppeteer to the code base. It is a headless browser that has a nifty screenshot function, which is used here to create images for the workbooks. All workbook-related functions have been moved to their own new module `workbook_generators.js`. It has two public-facing functions:
 - image.generate() (new!)
 - content.generate() (new fcn, pre-existing code moved from elsewhere)

When `generate_learning_path_markdown.js` is run, TWO new directories are created
 - workbooks (new!)
 - newsite (pre-existing, unchanged)

The `workbooks` folder is created by `image.generate()` function and contains all the workbook images together. Since it only contains images, it is copied to the website's `/static/images/learn/*` (instead of sharing `/content/en/learn/*` with contents of `newsite`).

Please let me know your questions of feedbacks, I'm happy to make changes. TY!


rrrutledge:
🎉for this work, @marshmallowrobot ❗️ I will take a look likely sometime next week.

rrrutledge:
OK so you're generating images on-the-fly when building the website?  Cool!  It's neat that you figured out not only how to make the images but how to make it easy for the next person to do the same. 

I am thinking that it would be good to adjust so that the image is created once manually and then checked in to the repo.  You can still check in the script for generating the image again if ever needed.   Update `/scripts/README.md` with instructions of how to to do so.

Great work ‼️

rrrutledge:
Oh - sorry that some of the work wasn’t used, but this approach does strike me as simpler to maintain.  Thank you for updating the image generation to be just HTML where it can be done in a web browser instead of the old way, which was some kind of image creation tool.

PR #523, opened 2023-01-13
Spellchecker for English LP content

spier:
I added a basic setup for a GitHub Action that runs a spellchecker for the LP content.
This is modelled after a very similar approach that we are just introducing for our Patterns (see https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/519).

A key difference here is that the tool used here, pyspelling, does seem to work a bit better with markdown, than it works with asciidocs. However I tweaked the configuration based on [an example](https://github.com/devonfw/hangar/blob/master/.spellcheck.yml) that I found, and it looks like it works well enough.

To judge the effect of the spellchecker, please compare the results of the [1st run](https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/runs/3915131601/jobs/6692970788), with the [2nd run](https://github.com/InnerSourceCommons/InnerSourceLearningPath/actions/runs/3915158590/jobs/6693033888). Before the 2nd run I had already fixed a couple of obvious spelling errors.

Left to do:

- decide if you like this approach
- decide if you want to spellcheck for en_US (so American English). if yes, remove the last block in `wordlist.txt` and fix the respective spellings (mostly just the word "organisation" 😄)
- improve the config in `.pyspelling.yml` and the `wordlist.txt`, so that the spellchecker passes on the current content (adapting the failing parts of the content is of course also another way to fix this)
- decide where to place the config files - they could be placed in some folder, if you don't want them in your root

I am happy to answer your questions. Just note, that I am only using pyspelling/aspell for a couple of days myself now :)

### DIFF
--- a/.github/workflows/spellcheck.yml
+++ b/.github/workflows/spellcheck.yml
@@ -1,0 +1,23 @@
+# Uses pyspelling (Aspell).
+# Further info in `.pyspelling.yml`
+name: Spell Check
+on:
+  push:
+    branches:
+    - main
+    paths:
+    - '**.asciidoc'
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  spellcheck:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Check Spelling
+      uses: rojopolis/spellcheck-github-actions@0.29.0
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ scripts/node_modules
 scripts/learningpath
 scripts/newsite
 scripts/.env
+wordlist.dic
--- a/.pyspelling.yml
+++ b/.pyspelling.yml
@@ -1,0 +1,41 @@
+# The exceptions below related to asciidocs where adapted from here:
+# https://github.com/devonfw/hangar/blob/master/.spellcheck.yml
+matrix:
+- name: Markdown
+  aspell:
+    lang: en
+    d: en_US
+  dictionary:
+    wordlists:
+    - wordlist.txt
+    output: wordlist.dic
+    encoding: utf-8
+  pipeline:
+  - pyspelling.filters.markdown:
+  - pyspelling.filters.html:
+      comments: true
+      attributes:
+      - title
+      - alt
+      ignores:
+      - ':matches(code, pre)'
+      - 'code'
+      - 'pre'
+  # Ignore for URL starting with http or https
+  - pyspelling.filters.context: 
+      context_visible_first: true
+      delimiters:
+      - open: 'https?:\S+?'
+        close: '\s'
+  # Ignore Image tags
+  - pyspelling.filters.context: 
+      context_visible_first: true
+      delimiters:
+      - open: 'image:[^[]+'
+        close: '\['
+  sources:
+  - 'introduction/[0-9][0-9]*.asciidoc'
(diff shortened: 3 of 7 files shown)


rrrutledge:
This is great, @spier ❗️ Thanks for this work.  Very excited to take a look.

rrrutledge:
This is great!  Where does the output go after the job is run?

spier:
@lenucksi the spell checker does not make any fixes itself. Not sure if it can be configured to do that.

What it does is that it shows the spelling issues in the output of the GitHub Action.
(@rrrutledge not sure if this answers your question about where the output goes?)

It is then up to the author to decide how to fix the issue:
- fix the words that the spell checker identified as problematic
- add the given word to the list of exceptions

FYI for our patterns repo I am experimenting with another spell checker right now ([Vale](https://vale.sh/)).
That one would show the spelling issues inline in the files. I can also do other checks for prose to get to a more consistent editorial style (inclusive language etc).

I am not sure yet which spell checker we will go with for our patterns repo.

So if you like, we can move this PR here back to Draft for the moment. Up to you.

rrrutledge:
This is very 🆒 , @spier .  I will bring this up in our next working group meeting.

> I am not sure yet which spell checker we will go with for our patterns repo.

I think that there's value in being on the same one, so let's see what you come up with.

> show the spelling issues inline in the files

As comments or something?  That would be great.

> spelling issues in the output of the GitHub Action

I see it now.  Great.

rrrutledge:
@spier is there an issue or a pull request that we can follow in the _InnerSourcePatterns_ repo for us to stay in touch with what spellchecker you are choosing?  We are happy to follow your lead, here.  Thanks for including us :).

spier:
@rrrutledge good idea. here is the issue: https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/519

rrrutledge:
Thank you, @spier ❗️ I added myself as a watcher.

rrrutledge:
We should update our `CONTRIBUTING.md` document to reference how to find and fix spelling errors in pull request.

rrrutledge:
Thanks for this work and this update, @spier !  We are very thankful and excited for spell checking.  Let us know if you have a rough idea of when you might work on it.  Any schedule is OK, but if you think that it will be a while before working on it some more then we may merge what's here so that at least we have _some_ spell checking in the meantime until your fancy, new Vale work is finished.

spier:
That is good clarification @rrrutledge. 

From previous comments I thought that the current solution would not be that useful for the LP content, which was one of the reasons for trying out vale.

So I will keep the current solution as a fallback option.

As for an estimate, I hope to have time around the Easter days, so within the next 10 days or so should be realistic.

rrrutledge:
> next 10 days or so 
Oh!  That’s not so long - we’ll probably wait.  We’d rather be in the same thing as InnerSourcePatterns.

rrrutledge:
@spier we saw that you created https://github.com/InnerSourceCommons/isc-styles for some more Vale work - so it looks like this may be moving along?  That's great!

spier:
@rrrutledge hehe you got your eyes everywhere :)

Yeah, I am trying to package up the styles/checks in a central place, so that it is easier to create some level of consistency across various ISC material.

I am not 100% there yet but I have a basic idea of how it might work. So yeah, this is moving along, slowly but surely :)

spier:
Closing this in favor of #560

PR #531, opened 2023-02-01
Create adding-trusted-committers.md

rrrutledge:
View [the document](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/rrrutledge-patch-2/docs/adding-trusted-committers.md).

### DIFF
--- a/docs/adding-trusted-committers.md
+++ b/docs/adding-trusted-committers.md
@@ -3,7 +3,7 @@ This document describes how to add a new Trusted Committer to the InnerSource Le
 1. Share your suggestion for new Trusted Commiter in the [:lock:learning-path-trusted-committers][learning-path-trusted-committers] private _Slack_ channel.
 2. If you have some agreeement (or even [lazy consensus](https://community.apache.org/committers/lazyConsensus.html)), then move forward.
 3. Create a pull request to the [Trusted Committers] section of the README with the new person.
-4. Invite the person into the [#learning-path-core] and [:lock:learning-path-trusted-committers][learning-path-trusted-committers] _Slack_ channels.
+4. Invite the person into the [:lock:learning-path-trusted-committers][learning-path-trusted-committers] _Slack_ channels.
 5. Add the person to the [@InnerSourceCommons/learning-path] _GitHub_ team.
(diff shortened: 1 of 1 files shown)


docs/adding-trusted-committers.md
```diff
@@ -0,0 +1,14 @@
+This document describes how to add a new Trusted Committer to the InnerSource Learning Path.
+
+1. Share your suggestion for new Trusted Commiter in the [:lock:learning-path-trusted-committers][learning-path-trusted-committers] private _Slack_ channel.
+2. If people seem to agree with your suggestion, then move forward.
+3. Create a pull request to the [Trusted Committers] section of the README with the new person.
+4. Invite the person into the [#learning-path-core] and [:lock:learning-path-trusted-committers][learning-path-trusted-committers] _Slack_ channels.
```
tsadler1988, line 6: > #learning-path-core

Are we making use of this, or does the private TCs channel supersede it?

docs/adding-trusted-committers.md
```diff
@@ -0,0 +1,14 @@
+This document describes how to add a new Trusted Committer to the InnerSource Learning Path.
+
+1. Share your suggestion for new Trusted Commiter in the [:lock:learning-path-trusted-committers][learning-path-trusted-committers] private _Slack_ channel.
+2. If people seem to agree with your suggestion, then move forward.
```
tsadler1988, line 4: Could maybe explain this a bit more e.g.

```suggestion
2. If you have '[lazy consensus](https://community.apache.org/committers/lazyConsensus.html)' (e.g. 1 or more approval and 0 rejections within ~1 week of suggesting), then move forward.
```

docs/adding-trusted-committers.md
rrrutledge: I think we could delete the “-core” version.

tsadler1988: > Does it really need the closed version?

This is potentially a bit sensitive. On the one hand, 'rejecting' a candidate TC, or having some other difficult discussion publicly, might not be good for psychological safety...but maybe having those discussions in the open will force people to be constructive with feedback e.g. "You're not quite ready for TC - here's what we'd like to see before granting you TC".

tsadler1988 (approved):

PR #520, opened 2022-12-07
Draft of the fulltext for the segment for product people

MaineC:
This builds on the detailed outline developed in #508 

It provides a fulltext version of a learning path segment for people experienced with product development. It tries to focus on the contributions that people who are not writing source code themselves can have on InnerSource projects.

PR #542, opened 2023-03-28
Review Product People Section

rrrutledge:
* Thank you for this work, @MaineC!
* Fixed some typos.
* @MaineC what future changes do you have planned?
* I think that, in-general, these articles could be split up so that each is more targeted.
With more targeted segments we may find there's enough overlap to merge some articles here into existing Learning Path segments or patterns and just link to them.
* We should decide on what checkpoint/level would we like for this work to make it to [the website](https://innersourcecommons.org/learn/learning-path/).

### DIFF
--- a/product/01-introduction.md
+++ b/product/01-introduction.md
@@ -1,10 +1,10 @@
 Welcome to the learning path section for product managers.
 
-After watching this segment, you will have a better understanding of how InnerSource speeds up product development.
+After completing this segment, you will have a better understanding of how InnerSource speeds up product development.
 We will also cover how it relates to Agile development best practices.
 
-To achive agility organizations strive for autonomous teams.
-However in a complex, interconnected world some dependencies cannot be avoided.
+To achieve agility, organizations strive for autonomous teams.
+However, in a complex, interconnected world, some dependencies cannot be avoided.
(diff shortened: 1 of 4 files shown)


MaineC (approved):

MaineC:
> what future changes do you have planned?

At some point I would like to address some of the open points that I mentioned at the original pull request, like https://github.com/InnerSourceCommons/InnerSourceLearningPath/pull/520#discussion_r1064174356 - however right now I'm lacking the time to thoroughly add those points in.

> With more targeted segments we may find there's enough overlap to merge some articles here into existing Learning 
> Path segments or patterns and just link to them.

That may well be possible.

One goal I had in mind with creating this segment (and I believe, there's still room for improvement): I wanted to create a segment, that people who have no background using version control can watch and understand. I also wanted to focus on providing input where InnerSource touches their work - and where their skills can help InnerSource projects.

The Introduction segment that we have to me seems a bit too introductory for someone who wants to dig deeper and understand the big picture. However the contributor and trusted committer segments have too many pieces that focus on the mechanics of coding. Then we have the product owner segment - but product owners are a role specific to how teams use specific agile methodologies... 

> We should decide on what checkpoint/level would we like for this work to make it to [the website](https://innersourcecommons.org/learn/learning-path/).

That is a good question indeed. In particular as when I tested the content one learning was that for the target group I have in mind, video snippets + workbook will likely not be sufficient. Interactivity and being able to directly put the learnings in context proved very helpful. Also there were many more and more diverse questions than what I'm used to from trainings targeting people with some level of software engineering background.

Issue #536, opened 2023-03-01
yasakawa user tile should link to github profile


rrrutledge:
Visit https://innersourcecommons.org/ja/learn/learning-path/introduction/01/

For some reason the `yasakawa` user tile links back to the same  learning path page.  It should link to https://github.com/yasakawa.

PR #545, opened 2023-03-29
Fix user tile bug

rrrutledge:
* For some reason, some author objects in the graph API are missing a user object (and subsequently a user URL).
* Users without a URL don't get a link to their GitHub profile.
* Fix is to explicitly ensure that we don't overwrite a good user URL with an empty one.
* See https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/536

### DIFF
--- a/scripts/get_contributors.js
+++ b/scripts/get_contributors.js
@@ -92,6 +92,8 @@ module.exports = async function (filepath) {
   return Object.values(
     [...authors, ...reviewers].reduce((accumulator, user) => {
       // Dedupe users
+      // Some user objects do not have a URL (can't figure out why).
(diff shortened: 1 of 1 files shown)


rrrutledge:
A better fix would be to figure out why some author objects in the API are getting returned without a user, but I couldn't see how to make much headway, there.

nysenthil (approved):
This morning, I discussed this PR with Russ and he explained his code changes. His tests ran fine as well. I'm approving this change.

nysenthil:
This morning, I discussed this PR with Russ to go over his code changes. All looks good. I'm approving and merging this PR now.

rrrutledge:
Looks good!

PR #547, opened 2023-04-03
Update CONTRIBUTING.md

rrrutledge:


rrrutledge (commented):
Heard from @mishari in _Slack_ that these instructions make sense.

PR #551, opened 2023-04-19
Documenting guidelines for translations

rrrutledge:
* Start with IBM Watson machine translation.
* Get a native speaker review

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ Here's a few tips as you go along:
 
 * All articles may be translated into any language of interest.
 * Instead of translating from scratch, consider contacting [@nysenthil] to get an IBM Watson machine translation as a starting point and revising from there.
-* All translations (manual or machine-generated) be reviewed and approved by a native speaker.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -14,8 +14,8 @@ Here's a few tips as you go along:
 # Translations
 
 * All articles may be translated into any language of interest.
-* Instead of translating from scratch, consider starting with [machine-translated text](https://www.deepl.com/translator) and revising from there.
-Remove "Translated by DeepL" in the final text.
+* Instead of translating from scratch, contact [@nysenthil] to get an IBM Watson machine translation as a starting point and revising from there.
```
tsadler1988, line 17:
```suggestion
* Instead of translating from scratch, consider contacting [@nysenthil] to get an IBM Watson machine translation as a starting point and revising from there.
```

I think retaining the word 'consider' is good to make it clear this is optional - maybe I'm overthinking it, but a potential contributor may be put off thinking they aren't allowed to manually translate from scratch.

rrrutledge: Good idea

PR #564, opened 2023-06-15
How to add a video to the Learning Path

rrrutledge:


docs/upload-new-video.md
```diff
@@ -0,0 +1,10 @@
+# How to upload a video for a new segment
+
+1.  Create the video file.
+2.  Create the [thumbnail].
+3.  Upload to [InnerSource Commons YouTube].
```
spier, line 5: Who to contact to get access to this channel?
Or is this documented elsewhere?

marshmallowrobot: Hey @spier!

As far as _who_ to contact for access... I believe any Trusted Committer should be able to grant access to the YouTube channel. I don't think this is officially documented anywhere. Even if there are cases where the listed TC can't directly grant access, they should have the contacts to get the request to someone that can.

It appears that different groups within InnerSource Commons maintain their own lists of Trusted Committers. Here are a couple of examples:
* [Patterns Trusted Commiters](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/TRUSTED-COMMITTERS.md#current-trusted-committers)
* [Learning Path Trusted Commiters](https://github.com/InnerSourceCommons/InnerSourceLearningPath/blob/main/README.md#trusted-committers)

Does this help?

marshmallowrobot (commented):
@rrrutledge This makes it sound so easy to add a new video! 

Copy-edit-wise: Please capitalize the title.

Content-wise, two things:
1.  As the resident subtitler, I would really love it if we asked content creators to add their own subtitles to their videos. It is my hope that this will result in videos where the host enunciates, and speaks more slowly. What do you think? If you agree, could you maybe add that as a blurb to the end of the list? 
2.  Regarding @spier's point about granting access to the channel, could that be added somewhere too?

rrrutledge:
Thanks, @marshmallowrobot !  Will take a look.

rrrutledge:
OK @marshmallowrobot take a look.

Issue #556, opened 2023-04-29
github-actions bot shouldn't be listed as a contributor


rrrutledge:
Now that the bot is [checking in updated links](https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/538) it is listed as a contributor in the "Acknowledgements" section at the bottom of generated articles.

We should only acknowledge human contributors.

PR #565, opened 2023-06-26
Bot user should not show as contributor

rrrutledge:
* Filter bot users out of contributors list.
* You can see the bot user missing from the list of contributors on the (local) left article with these changes, whereas it is present on the right that does not have these changes.  https://innersourcecommons.org/it/learn/learning-path/trusted-committer/01/
* Resolves https://github.com/InnerSourceCommons/InnerSourceLearningPath/issues/556.

image

### DIFF
--- a/scripts/get_contributors.js
+++ b/scripts/get_contributors.js
@@ -89,14 +89,18 @@ module.exports = async function (filepath) {
     })
   })
 
+  // Dedupe users
   return Object.values(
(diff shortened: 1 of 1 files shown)


nysenthil (approved):
Hi @rrrutledge 
Your code change looks good with the new conditional if statement to add the new user only when the user name doesn't include a substring [bot] in it. This change can be merged now.

PR #574, opened 2023-07-31
Adding a metaphor of hammers as tools used in silos

mishari:
Adding an outline


rrrutledge (approved):
Yes!  Very much looking forward to this!

rrrutledge (approved):

PR #583, opened 2023-08-15
Added concurrency section to Publish Website workflow (Issue 575)

marshmallowrobot:
When builds are triggered within about 4 min of each other, the parallel calls to GitHub's GraphQL API clobber one another. Even with a throttling helper, the clobbering results in a too-many-requests API error, and the build fails.

This small change should hopefully prevent this error in the future. It will check if the build/workflow is already running, and if so, cancels it. If there are multiple builds in quick succession, only the latest one will publish.

### DIFF
--- a/.github/workflows/publish-to-website.yml
+++ b/.github/workflows/publish-to-website.yml
@@ -14,6 +14,11 @@ on:
       - trusted-committer/**
       - workbook/**
 
+# Checks if this workflow is already running, and if so cancels it
+concurrency:
+  group: ${{ github.workflow }}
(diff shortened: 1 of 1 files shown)


tsadler1988 (approved):
Great catch!

tsadler1988:
Docs for reference: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

rrrutledge:
Great work!

PR #597, opened 2025-01-29
Fixing vale setup

spier:
We noticed over at https://github.com/InnerSourceCommons/managing-innersource-projects/pull/91 that the vale setup with custom vocabulary files did not work anymore.

The issue was related to a change in the required folder structure for custom vocabulary files. See [docs](https://vale.sh/docs/keys/vocab#folder-structure).

Specifics:

```
In versions of Vale prior to 3.0, vocabularies were stored in <StylesPath>/Vocab.
When upgrading from an older version of Vale, you'll need to move your vocabularies to the new <StylesPath>/config/vocabularies location.
```

While fixing this, I also made the GH actions that we use in our workflow more stable.


rrrutledge (approved):
Thank you!

